PR #2, opened 2018-03-12
enable AppVeyor CI

dirk-thomas:


test/test_executor_sequential.py
```diff
@@ -95,6 +101,8 @@ def delayed_sigint():
         os.kill(
             os.getpid(),
             signal.SIGINT if sys.platform != 'win32' else signal.CTRL_C_EVENT)
+        if sys.platform == 'win32':
+            os.kill(signal.CTRL_C_EVENT)
```
dirk-thomas: Missing first PID argument. See #309.

PR #16, opened 2018-04-18
use rc 0 when pytest has failing tests

dirk-thomas:
`pytest` returns different return codes for several causes (see https://docs.pytest.org/en/latest/usage.html#possible-exit-codes). When any of the tests failed that should not be considered an "error" in terms of invoking the `test` verb. This unifies the return value semantic across other types.

Related to colcon/colcon-cmake#3.


mikaelarguedas:
same question as https://github.com/colcon/colcon-cmake/pull/3#issuecomment-382572995

mikaelarguedas:
Feature addition ticketed at colcon/colcon-core#21

mikaelarguedas:
@dirk-thomas this seems to still fail sometimes: https://ci.ros2.org/job/ci_osx/3517

PR #17, opened 2018-04-18
enumerate duplicate packages in exception

dirk-thomas:
To provide useful feedback to the user in case this happens.

### DIFF
--- a/colcon_core/package_selection/__init__.py
+++ b/colcon_core/package_selection/__init__.py
@@ -1,6 +1,7 @@
 # Copyright 2016-2018 Dirk Thomas
 # Licensed under the Apache License, Version 2.0
 
+from collections import defaultdict
 import traceback
 
 from colcon_core.logging import colcon_logger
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
thanks for tackling it I didnt have time to submit the PR yet

mikaelarguedas:
Are you also working on changing the default source path or should I submit a PR for that ?

dirk-thomas:
> Are you also working on changing the default source path

I would rather keep it as it is (which is the current directory). The `src` directory is a pretty ROS specific thing. It should still work and find the package in the `src` directory though.

mikaelarguedas:
> I would rather keep it as it is (which is the current directory). The src directory is a pretty ROS specific thing. It should still work and find the package in the src directory though.

We should then make a note for ament users to either nuke their build/install folders or to specify the source path when running colcon on an ament workspace for the first time.

Context: ament doesn't drop AMENT_IGNOREs in the build/install space. So running colcon at the root of an ament workspace will complain about duplicated packages (that will have more meaningful error message once this PR is validated and integrated).

dirk-thomas:
> ament doesn't drop AMENT_IGNOREs in the build/install space

Can you double check the build space. The `ament_tools` code indicates it does: https://github.com/ament/ament_tools/blob/e978d4ff07f2c4e1a59a58429a6b8f48cd730715/ament_tools/verbs/build_pkg/cli.py#L321-L326

For the install space we could add a check for `local_setup.sh|bat` and `_order_isolated_packages.py`. If both of these files or in a given directory then conclude that it is an ament install space and ignore it. Does that sound like a good approach?

mikaelarguedas:
> Can you double check the build space. The ament_tools code indicates it does

I meant the root of the build space. You're right that we added AMENT_IGNORE in package specific in the build space so this shouldn't be an issue :+1:

mikaelarguedas:
> For the install space we could add a check for local_setup.sh|bat and _order_isolated_packages.py. If both of these files or in a given directory then conclude that it is an ament install space and ignore it. Does that sound like a good approach?

Yeah that sounds like a good approach

PR #24, opened 2018-04-23
modify destination collector to store option as well as destination

dirk-thomas:
Related to colcon/colcon.readthedocs.org#5.

This patch modifies the destination collector to save the name of the command line options as well as the destination. That allows that configuration files can use the same spelling as the command line options (as well as e.g. a short version of the argument) while the internal logic uses the destination name for valid variables names.

### DIFF
--- a/colcon_core/argument_parser/destination_collector.py
+++ b/colcon_core/argument_parser/destination_collector.py
@@ -1,11 +1,13 @@
 # Copyright 2016-2018 Dirk Thomas
 # Licensed under the Apache License, Version 2.0
 
+from collections import OrderedDict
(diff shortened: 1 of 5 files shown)


colcon_core/argument_parser/destination_collector.py
```diff
@@ -26,16 +28,21 @@ def get_destinations(self):
         :returns: The destination names
         :rtype: list
```
dirk-thomas: The return type was fixed in a follow up commit: 0a844f741dd83c5d6cdfd18ed2ae685427a7b62a

PR #25, opened 2018-04-24
add CommandEnded to log return code of each invoked command

dirk-thomas:
This should be helpful in debugging the return code of each invoked command.

~~Before this should be merged the test needs to be updated to cover the new event class.~~ I updated the test to cover the new event class.

### DIFF
--- a/colcon_core/event/command.py
+++ b/colcon_core/event/command.py
@@ -34,7 +34,11 @@ class Command:
         environment variable and the command including the arguments.
         """
         string = "Invoking command in '{self.cwd}': ".format_map(locals())
+        string += self._get_env_string()
+        string += self._get_cmd_string()
+        return string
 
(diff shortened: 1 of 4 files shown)


dirk-thomas:
> Is there a job using this (showing the new output)?

No, the output only affects the log files. But I have tested the patch locally.

dirk-thomas:
> Is the goal that now if a job crash on CI we can get the exact return code from the log files in the workspace?

Yes.

> Is there a way to retrieve this information from a job if the workspace has been overwritten due to another job being started on the same node?

No, the output is not printed to the console. This is basically the continuous balance between more verbose and overloaded. Currently `colcon` only prints "start" / "stop" / "summary" to the console as well as the output of the actual command being run. The log files contain all additional information like "what command was run", etc.

PR #29, opened 2018-04-25
add a log file containing all log messages

dirk-thomas:
This provide more debugging information even in the case where the tool was invoked without any custom log level.


mikaelarguedas:
:+1: thanks for confirming

PR #31, opened 2018-04-26
add verb specific latest symlink

dirk-thomas:
Beside the `latest` symlink` this patch introduces a `latest_{verb}` symlink to easily get the latest `build` / `test` logs.

### DIFF
--- a/colcon_core/location.py
+++ b/colcon_core/location.py
@@ -109,12 +109,16 @@ def set_default_log_path(*, base_path, env_var=None, subdirectory=None):
 _create_log_path_lock = Lock()
 
 
-def create_log_path():
+def create_log_path(verb_name):
     """
     Create the logging directory if it doesn't exist yet.
 
-    On non-Windows platforms a symlink named `latest` is created in the base
-    path which links to the subdirectory.
+    On non-Windows platforms two symlinks are created as siblings of the log
+    path:
+    * `latest_<verb_name>` linking to the log path
+    * `latest` linking to `latest_<verb_name>`
+
+    :param str verb_name: The verb name
     """
     path = get_log_path()
     if path.exists():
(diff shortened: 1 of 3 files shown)


colcon_core/location.py
```diff
@@ -130,23 +134,33 @@ def create_log_path():
         # create latest symlink
         if sys.platform == 'win32':
             return
```
mikaelarguedas: Nitpick: Not sure if it's worth checking that twice. I guess you want to check it in the function itself in case you want to use it in other places.

colcon_core/location.py
```diff
@@ -109,12 +109,16 @@ def set_default_log_path(*, base_path, env_var=None, subdirectory=None):
...
     Create the logging directory if it doesn't exist yet.
 
-    On non-Windows platforms a symlink named `latest` is created in the base
-    path which links to the subdirectory.
+    On non-Windows platforms two symlinks are created as siblings of the log
+    path:
+    * `latest_<verb_name>` linking to the log path
+    * `latest` linking to `latest_<verb_name>`
+
+    :param str verb_name: The verb name
```
mikaelarguedas: Question: This seem inconsistent across the project (even within this file), should the parameter type be specified within the `:param` tag or not?

It seems common to specify the type on another line using the `:type` tag

colcon_core/location.py
```diff
@@ -130,23 +134,33 @@ def create_log_path():
         # create latest symlink
```
mikaelarguedas: Nit: create latest symlink**s**

colcon_core/location.py
dirk-thomas: I removed the check from the function. Since it is private it is not supposed to be used from other places anyway.

colcon_core/location.py
dirk-thomas: This file looks consistent to me. I assume you refer the use of `:rtype:`? For parameters the type can be mentioned between the keyword and the parameter name. I don't think that is possible for return values.

colcon_core/location.py
dirk-thomas: Fixed in 584206d2ef316739e5f34d7279fc9efe1592f812.

colcon_core/location.py
mikaelarguedas: I was refering to https://github.com/colcon/colcon-core/blob/584206d2ef316739e5f34d7279fc9efe1592f812/colcon_core/location.py#L49 for example

dirk-thomas: In that case the argument is not a "simple" builtin type but an instance of a specific class. I often don't put the whole type name in there since it makes the part super long and I haven't experimented with the API extraction yet.

mikaelarguedas: both 
```
:param str verb_name: The verb name
```
and
```
:param verb_name: The verb name
:type verb_name: str
```
are valid in sphinx, it's just a matter of what convention we chose to follow.

dirk-thomas: Sure, as I mentioned almost all of the code uses the first option to avoid the duplication.

dirk-thomas: Sure, as I mentioned almost all of the code uses the first option to avoid the duplication. But not all cases specify the type if it is not a "primitive" type.

mikaelarguedas: sounds good :+1:

PR #38, opened 2018-04-27
add setup script, rename prefix script into local_setup

dirk-thomas:
While not complete (the `.bat` scripts still need to be refactored) I would like to get a first review / feedback if that is the desired result.

### DIFF
--- a/colcon_core/shell/__init__.py
+++ b/colcon_core/shell/__init__.py
@@ -310,3 +310,39 @@ def create_environment_hook(
             'Could not find a primary shell extension for creating an '
             'environment hook')
     return hooks
+
+
+_get_colcon_prefix_path_warnings = set()
+
+
+def get_colcon_prefix_path(*, skip):
+    """
+    Get the paths from the COLCON_PREFIX_PATH environment variable.
+
+    For not existing paths a warning is being printed and the path is being
+    skipped.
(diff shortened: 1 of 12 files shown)


colcon_core/shell/__init__.py
```diff
@@ -310,3 +310,40 @@ def create_environment_hook(
             'Could not find a primary shell extension for creating an '
             'environment hook')
     return hooks
+
+
+_colcon_prefix_path_cache = {}
```
sloretz: Why cache the result? Can the `COLCON_PREFIX_PATH` environment variable change between calls to `get_colcon_prefix_path()`?

dirk-thomas: I tried to describe that in the docblock:

> The result is being cached based on the passed skip parameter to avoid printing the warning on repeated invocations.

While the environment isn't supposed to change during one run of the tool the fact that the result would be invalid if the environment does change isn't really a good idea.

I refactored the logic in 33b1d05e9ced08f109ad31ecb13ad1c8b79b50c9 to actually store which warning has already been logged and just use that global information to avoid repeating the warning. The computation happens every time and is therefore correct no matter what happens to the environment.

dirk-thomas:

PR #41, opened 2018-05-03
add test failure event

dirk-thomas:
This will allow to react to failed tests. E.g. as also part of this patch mention the test failures as part of the "Finished <<< {pkgname}" message.


dirk-thomas:

PR #43, opened 2018-05-04
warn about duplicate entry points

dirk-thomas:


dirk-thomas:
Ignoring failing CI which is due to a regression in the latest `flake8-builtins` release.

PR #44, opened 2018-05-09
rename SkipLocationException to IgnoreLocationException

dirk-thomas:
Since the semantic of "skip" and "ignore" is different within the command line arguments the exception should follow the same meaning in order to not confuse the developer.


dirk-thomas:

dirk-thomas:
Addressed some style issues in 89edc0fb605edd8980fd1d4c25c6fd5095cc6ab4.

PR #45, opened 2018-05-10
add stdeb.cfg, update dependencies

dirk-thomas:


dirk-thomas:

PR #46, opened 2018-05-10
ignore log directory

dirk-thomas:
To avoid crawling through all log directories when not search through `src` explicitly.


dirk-thomas:
Rebased - "fixed" the new CI failures.

dirk-thomas:

PR #47, opened 2018-05-10
Travis bootstrap needs pytest-repeat and pytest-rerunfailures

sloretz:
#46 bootstrap travis tests failed with `pkg_resources.DistributionNotFound: The 'pytest-rerunfailures' distribution was not found and is required by colcon-core`. This attempts to fix it.


dirk-thomas:
The regression is coming from #45. I guess that happens when you start ignoring CI failures... :disappointed:

The same will probably be needed for `pytest-repeat` too.

sloretz:
Only test failure is `AssertionError: flake8 reported 1875 errors` which is unrelated. Merging.

PR #63, opened 2018-05-31
override argument parser to fallback to args rather than options

### LINKED ISSUE #36: consider custom parsing logic to avoid leading space for nested arguments
Currently nested arguments like `--cmake-args` need to be prefixes with a space in order to not be considered by `argparse`. If that should be avoided custom argument parsing logic needs to be implemented which pre-processes the arguments to extract all following parameters up to a `--`.

dirk-thomas:
Fixes #36.

This is an "interesting" approach to monkey patch the change proposed upstream.

I will put this in review for people to try and comment on the patch. Once it is approved the help texts can be updated to mention the space prefix only for cases the the arguments are actually valid options of `colcon`.

### DIFF
--- a/colcon_core/command.py
+++ b/colcon_core/command.py
@@ -139,8 +139,21 @@ def create_parser(environment_variables_group_name):
       the environment variable extensions
     :returns: The argument parser
     """
+    # workaround a limitation in argparse to accept arguments to options
+    # which begin with a dash but are not options themselves
+    # https://bugs.python.org/issue9334
+    class CustomArgumentParser(argparse.ArgumentParser):
+
+        def _parse_optional(self, arg_string):
+            result = super()._parse_optional(arg_string)
+            if result == (None, arg_string, None):
+                # in the case there the arg is classified as an unknown 'O'
+                # override that and classify it as an 'A'
+                return None
+            return result
+
     # top level parser
-    parser = argparse.ArgumentParser(
+    parser = CustomArgumentParser(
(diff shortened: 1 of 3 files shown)


dirk-thomas:
While I picked @mikaelarguedas as the reviewer I hope others will give it a try to and review the patch to move forward on this quickly.

dirk-thomas:
> For example, here I was expecting a symlink install:

It seemed to work for options following the greedy one in my local testing. Can you double check the logs what args have been parsed.

> Is there some kind of delimimter that should be used ?

No delimiter.

dirk-thomas:
Also console bridge is a plain CMake package, no? I doubt that can do a symlink install in the first place...

mikaelarguedas:
Oh yeah that could be it, I took the first non-python package from my topological order ^^ poor choice...

Works fine with class_loader :+1:  sorry for the noise

```
-- Symlinking: /home/mikael/work/ros2/bouncy_ws/install/class_loader/include/class_loader/multi_library_class_loader.hpp
-- Symlinking: /home/mikael/work/ros2/bouncy_ws/install/class_loader/include/class_loader/register_macro.hpp
-- Symlinking: /home/mikael/work/ros2/bouncy_ws/install/class_loader/include/class_loader/visibility_control.hpp
```

mikaelarguedas:
I didn't get a chance to test many configurations but, from the few scenarios I tested, this PR works perfectly :+1: 
(tested only `cmake-args` and `ctest-args` so far)

PR #69, opened 2018-06-11
declare flake8 import order style

dirk-thomas:
To avoid relying on other (potentially user specific) configurations.


dirk-thomas:
I would interpret a review of this PR as a "wildcard approval" for all the referenced PRs applying the same change.

PR #70, opened 2018-06-13
create file with run deps for each package, use these for sourcing packages in topological order

### LINKED ISSUE #56: support installing to same prefix from different workspaces
Currently the prefix-level scripts contain a list of packages in the workspace. The advantage of that approach is that no crawling of the file system is necessary when sourcing the script.

In order to support installing into the same prefix from multiple workspaces the list of packages needs to be collected by crawling the file system, ordering the packages topologically and sourcing each package. Parsing the package manifests will not be sufficient since some package might not have one and their dependencies were determined using external information which is not available in the install prefix anymore.

dirk-thomas:
Addresses #56.

Since the `pkg_names` are not necessary anymore in the `create_prefix_script` function of the `ShellExtensionPoint` I removed the argument from the signature. As a consequence I bumped the extension version accordingly (which is the first time) and its nice to see the extension point versioning mechanism working in action :smirk:

### DIFF
--- a/colcon_core/shell/template/prefix.bat.em
+++ b/colcon_core/shell/template/prefix.bat.em
@@ -6,13 +6,17 @@
 
 :: add this prefix to the COLCON_PREFIX_PATH
 call:_colcon_prefix_bat_prepend_unique_value COLCON_PREFIX_PATH "%%~dp0"
-@[if pkg_names]@
+
+:: get all packages in topological order
+call:ament_get_ordered_packages _ordered_packages "%~dp0"
 
 :: source packages
-@[  for pkg_name in pkg_names]@
-call:_colcon_prefix_bat_call_script "%%~dp0@('' if merge_install else (pkg_name + '\\'))share\@(pkg_name)\@(package_script_no_ext).bat"
-@[  end for]@
-@[end if]@
+if "%_ordered_packages%" NEQ "" (
+  for %%p in ("%_ordered_packages:;=";"%") do (
+    call:_colcon_prefix_bat_call_script "%%~dp0@('' if merge_install else '%%~p\\')share\%%~p\@(package_script_no_ext).bat"
+  )
+  set "_ordered_packages="
+)
 
 goto:eof
 
@@ -49,6 +53,36 @@ goto:eof
   :: set result variable in parent scope
   endlocal & (
     set "%~1=%all_values%"
+  )
+goto:eof
+
+
+:: Get the package names in topological order
+:: using semicolons as separators and avoiding leading separators.
+:: first argument: the name of the result variable
+:: second argument: the base path to look for packages
+:ament_get_ordered_packages
+  setlocal enabledelayedexpansion
+
+  :: use the Python executable known at configure time
+  set "_colcon_python_executable=@(python_executable)"
+  :: allow overriding it with a custom location
+  if "%COLCON_PYTHON_EXECUTABLE%" NEQ "" (
+    set "_colcon_python_executable=%COLCON_PYTHON_EXECUTABLE%"
+  )
+
+  set "_colcon_ordered_packages="
+  for /f %%p in ('""%_colcon_python_executable%" "%~dp0_local_setup_util.py"@
+@[if merge_install]@
+ --merged-install@
+@[end if]@
+"') do (
+    if "!_colcon_ordered_packages!" NEQ "" set "_colcon_ordered_packages=!_colcon_ordered_packages!;"
+    set "_colcon_ordered_packages=!_colcon_ordered_packages!%%p"
+  )
+  endlocal & (
+    :: set result variable in parent scope
+    set "%~1=%_colcon_ordered_packages%"
   )
 goto:eof
 
--- a/colcon_core/shell/template/prefix.bat.em
+++ b/colcon_core/shell/template/prefix.bat.em
@@ -13,7 +13,7 @@ call:ament_get_ordered_packages _ordered_packages "%~dp0"
 :: source packages
 if "%_ordered_packages%" NEQ "" (
   for %%p in ("%_ordered_packages:;=";"%") do (
-    call:_colcon_prefix_bat_call_script "%%~dp0@('' if merge_install else '%%~p\\')share\%%~p\@(package_script_no_ext).bat"
+    call:_colcon_prefix_bat_call_script "%~dp0@('' if merge_install else '%%~p\\')share\%%~p\@(package_script_no_ext).bat"
   )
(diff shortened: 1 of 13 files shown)


mikaelarguedas:
Does this change implies that it will not be possible to source a workspace on a system that doesn't have python ?

It would be great to fall back to pure sh / bash / bat in the case where Python is not available instead of exiting (if possible).
The use case I have in mind is embedded cases where the workspace will be compiled on a Python capable environment and then deployed on a trimmed down platform that doesn't have Python support.

dirk-thomas:
> Does this change implies that it will not be possible to source a workspace on a system that doesn't have python ?

Correct.

A shell-only implementation of the logic in the utilities module is certainly possible. I would need to be implementation for each primary shell type.

Please feel free to create a follow up ticket for this (which was also proposed for `ament_tools` in the past but was never implemented).

mikaelarguedas:
My understanding is that without this change, sourcing works on platforms that don't have python. Is that correct?

If the answer is yes I think we should not merge this if it doesn't at least fall back to the current implementation working without Python (but slower).

We can totally create a follow-up ticket to improve the logic to be more efficient in each shell type :+1:

dirk-thomas:
> If the answer is yes I think we should not merge this if it doesn't at least fall back to the current implementation working without Python (but slower).

I highly disagree with your rational. With `ament_tools` (which is currently the recommended build tool) the setup files *require* a Python module. This is *exactly* the same as with `colcon` with this patch applied. Blocking the transition from `ament_tools` to `colcon` for a feature requirement which wasn't even satisfied before doesn't sound reasonable to me.

--

If a user doesn't have Python available it is possible to workaround the limitation by sourcing the package specific setup files manually or setting environment variable manually all together (which is the same for `ament_tools` as well as `colcon`).

mikaelarguedas:
I acknowledge not having the full picture here.

From my perspective, this was blocking the embedded use case and was finally enabled by colcon, that is why I perceive it as a last minute regression that wasn't originally planned (in my mind) for Bouncy.

That being said, I do see the benefit of this set of PRs and that is why I suggested the compromise of falling back to current behavior.

> Blocking the transition from ament_tools to colcon for a feature requirement which wasn't even satisfied before doesn't sound reasonable to me.

I looks lilke that's the part I'm missing leading to the misunderstanding, why is this blocking the transition from `ament_tools` to `colcon`?

dirk-thomas:
> why is this blocking the transition from `ament_tools` to `colcon`?

Because users expect #56 to work and it would break if we roll out colcon as is.

dirk-thomas:
> Can you please open a follow-up ticket with a descriptin and an estimate of the work needed to implement a fallback to what colcon currently does. Hopefully we will be able to address it before the release.

I created #73 to capture the idea of Python-less prefix scripts. I don't think that an in-between solution makes any sense (as it is currently using a hard coded list of packages). Implementing the Python logic in shell script is pretty certain out-of-scope for the upcoming release.

dirk-thomas:
> Let's merge this as is then.

Can you please review this PR as well as the connected ones.

colcon_core/shell/template/prefix.bat.em
```diff
@@ -53,6 +57,36 @@ goto:eof
 goto:eof
 
 
+:: Get the package names in topological order
+:: using semicolons as separators and avoiding leading separators.
+:: first argument: the name of the result variable
+:: second argument: the base path to look for packages
+:ament_get_ordered_packages
```
dhood: is it a mistake that this is `ament`?

colcon_core/shell/template/prefix.bat.em
```diff
@@ -53,6 +57,36 @@ goto:eof
...
+
+  :: use the Python executable known at configure time
+  set "_colcon_python_executable=@(python_executable)"
+  :: allow overriding it with a custom location
+  if "%COLCON_PYTHON_EXECUTABLE%" NEQ "" (
+    set "_colcon_python_executable=%COLCON_PYTHON_EXECUTABLE%"
+  )
+
+  set "_colcon_ordered_packages="
+  for /f %%p in ('""%_colcon_python_executable%" "%~dp0_local_setup_util.py"@
```
dhood: there's a trailing @ here that might be an accident (or maybe it's just to cut off the newline even if there isn't any other empy stuff in this line): I just haven't seen it before so wanted to draw attention to it

colcon_core/shell/template/prefix.sh.em
```diff
@@ -63,7 +63,23 @@ _colcon_prefix_sh_prepend_unique_value() {
...
+
+# use the Python executable known at configure time
+_colcon_python_executable="@(python_executable)"
+# allow overriding it with a custom location
+if [ -n "$COLCON_PYTHON_EXECUTABLE" ]; then
+  _colcon_python_executable="$COLCON_PYTHON_EXECUTABLE"
+fi
+# if the Python executable doesn't exist try another fall back
+if [ ! -f "$_colcon_python_executable" ]; then
+  if /usr/bin/env python3 --version > /dev/null
```
dhood: has this been tested on osx yet? might have issues based on experience with https://github.com/ros2/ci/pull/127/files

colcon_core/shell/template/prefix.bat.em
dirk-thomas: That is intentionally. In an `empy` template this will suppress the following character which in this case is the newline. The result is that the optional `--merge-install` appears on the same line as the Python invocation. The `for` loop line extends from line 75 to line 79 in the template but results in a single line in the output.

colcon_core/shell/template/prefix.sh.em
dirk-thomas: Not specifically. It is only meant as a fallback.

colcon_core/shell/template/prefix.bat.em
dirk-thomas: Certainly a mistake - good catch! Fixed in 930136de6aed28004b6ea79f8982a99f3c014dc3.

colcon_core/shell/template/prefix.sh.em
dhood: yeah and the issue I referenced might only be a problem if SIP stripping the DYLD_LIBRARY_PATH is an issue, now that I look closer.. not something that needs to be worried about now

colcon_core/shell/template/prefix.bat.em
dhood: OK that's what I was thinking after a while but it's good to get clarification thanks!

PR #72, opened 2018-06-14
add pytest-runner as dependency

dirk-thomas:
This will address the problem with the Bouncy devel jobs not picking up the `pytest` tests. Basically without it `setup.py test` doesn't find any tests.

This requires `python3-pytest-runner` to be available in Ubuntu Xenial (on all other platforms it is available out-of-the-box). I tried installing the [Bionic Debian](https://launchpad.net/ubuntu/+archive/primary/+files/python3-pytest-runner_2.11.1-1_all.deb) in a Xenial docker which worked without a problem. @tfoote Can you please import this Debian into our bootstrap repo.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,6 +35,7 @@ install_requires =
   pytest-cov
   pytest-repeat
   pytest-rerunfailures
+  pytest-runner
   setuptools
 packages = find:
 tests_require =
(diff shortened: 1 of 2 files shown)


tfoote:
@dirk-thomas It's in the bootstrap repo and imported to build.ros2.org: http://build.ros2.org/job/import_upstream/33/ It should be available on all arches.

Issue #85, opened 2018-07-17
Better error msg if compile-time Python not found at sourcing time on Windows


mikaelarguedas:
Since https://github.com/colcon/colcon-core/pull/70 we use Python to source packages in topological order.
If the python executable is not found it crashes with `The system cannot find the path specified.`

We should provide a better error message so that users know what fails.

One way to do it might be to check that the file exist before calling it. But it may be to expensive...

dirk-thomas:
> One way to do it might be to check that the file exist before calling it. But it may be to expensive...

That sounds good and is certainly not too expensive. Each recusrively sourced script is checked for existence before so one extra check won't be significant. Please consider to create a pull request for the additional check.

PR #86, opened 2018-07-18
improve check and fallback behavior for python executable

mikaelarguedas:
FIxes https://github.com/colcon/colcon-core/issues/85

Opening for visibility needs some testing still

### DIFF
--- a/colcon_core/shell/template/prefix.bat.em
+++ b/colcon_core/shell/template/prefix.bat.em
@@ -78,7 +78,7 @@ goto:eof
     if not exist "%_colcon_python_executable%" (
       python --version > NUL 2> NUL
       if errorlevel 1 (
-        echo error: unable to find python3 executable
+        echo error: unable to find python executable
         exit /b 1
       )
       set "_colcon_python_executable=python"
--- a/colcon_core/shell/template/prefix.bat.em
+++ b/colcon_core/shell/template/prefix.bat.em
@@ -64,26 +64,24 @@ goto:eof
 :_colcon_get_ordered_packages
   setlocal enabledelayedexpansion
 
-  :: use the Python executable known at configure time
-  set "_colcon_python_executable=@(python_executable)"
-  :: allow overriding it with a custom location
+  :: check environment variable for custom Python executable
   if "%COLCON_PYTHON_EXECUTABLE%" NEQ "" (
     if not exist "%COLCON_PYTHON_EXECUTABLE%" (
-      echo error: Provided '%COLCON_PYTHON_EXECUTABLE%' does not exist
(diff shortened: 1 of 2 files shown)


colcon_core/shell/template/prefix.bat.em
```diff
@@ -70,6 +70,23 @@ goto:eof
   if "%COLCON_PYTHON_EXECUTABLE%" NEQ "" (
     set "_colcon_python_executable=%COLCON_PYTHON_EXECUTABLE%"
   )
+  if not exist "%_colcon_python_executable%" (
+    rem provided colcon_python_executable not found
+    rem falling back to build time python
+    if not exist "%(python_executable)" (
```
dirk-thomas: I would suggest not to check for the compile time `python_executable` again. That was already the starting value when not overridden with an environment variable. I would keep the logic as close as possible to what is done in `sh`: https://github.com/colcon/colcon-core/blob/aca10c55a6c7749777171bb18fce20ea27ca0d1d/colcon_core/shell/template/prefix.sh.em#L67-L82

mikaelarguedas: This means that if the user provides an invalid override we will never fall back to the python used at compile time.

Though it seems more likely that the one used for compiling will have more change to provide expected results than the one on the path that may be a different version (that could actually be python 2 on windows with the current patch).

Can you clarify why we should not fall back to the compile time one if available?

dirk-thomas: > Can you clarify why we should not fall back to the compile time one if available?

If the user provide a Python path it probably shouldn't even try any fall back. Because the user asked explicitly to use that. The `sh` logic could be changed to not try a fall back in that case.

dirk-thomas:
I updated the `sh` logic in 281839a612c0676a3d5c0865d1320fda7f3f267b:
* improve the error message mentioning the environment variable
* move variable closer to where they are actually used
* catch `stderr` output when checking for fallback executable
* invert the logic to get rid of the else block (to be shorter and the same as above)

colcon_core/shell/template/prefix.bat.em
```diff
@@ -64,11 +64,25 @@ goto:eof
...
+    )
     set "_colcon_python_executable=%COLCON_PYTHON_EXECUTABLE%"
+  ) else (
+    :: use the Python executable known at configure time
+    set "_colcon_python_executable=@(python_executable)"
+    :: if it doesn't exist try a fall back
+    if not exist "%_colcon_python_executable%" (
+      python --version > NUL 2> NUL
+      if errorlevel 1 (
+        echo error: unable to find python3 executable
```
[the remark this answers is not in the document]
mikaelarguedas: :+1: I'll go ahead and merge this and connected PRs then

PR #88, opened 2018-07-19
expose file extensions covered by a shell extension

dirk-thomas:
Making the file extensions for each shell extension available enables colcon/colcon-ros#29 which avoids hard coding those file extensions.

### DIFF
--- a/colcon_core/shell/__init__.py
+++ b/colcon_core/shell/__init__.py
@@ -55,6 +55,18 @@ class ShellExtensionPoint:
     """
     PRIORITY = 100
 
+    def get_file_extensions(self):
+        """
+        Get the file extensions provided by this extension.
+
+        By default the extension name will be returned.
+        The method is intended to be overridden in a subclass.
+
+        :returns: the file extensions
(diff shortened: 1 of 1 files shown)


colcon_core/shell/__init__.py
```diff
@@ -50,11 +50,23 @@ class ShellExtensionPoint:
...
+    All "non-primary" shell extensions must use a priority equal to or lower
     than the default.
     """
     PRIORITY = 100
 
+    def get_file_extensions(self):
+        """
+        Get the file extensions provided by this extension.
+
+        By default the extension name will be returned.
```
mikaelarguedas: By default the *shell name* will be returned ?

dirk-thomas: The name of the extension will be returned - as specified for the entry point.

mikaelarguedas: yeah it just seem more prone to confusion to refer to the "file extension" or the "colcon extension" that's why I suggested a different wording

PR #91, opened 2018-07-25
setuptools min version

### LINKED ISSUE #90: python3-colcon `colcon build` reports an error on Ubuntu Xenial
`colcon build` on an ubuntu:xenial container reports the following error:

```
ERROR:colcon.colcon_core.entry_point:Exception loading extension 'colcon_core.package_identification.python': No module named 'setuptools.config'
```

It looks like the system packaged `setuptools` doesn't include this module in xenial - this can be worked around by upgrading the system `setuptools` using `sudo python3 -m pip install setuptools -U`.

dirk-thomas:
Fixes #90.

The first commit declares the minimum version of setuptools required. This only works when using `pip` to install the package. The version dependency can't be specified in the `stdeb.cfg` file since some supported platforms like Ubuntu Xenial don't provide a new enough Debian package of `setuptools`.

The second commit will catch the case where the available `setuptools` version is too old and provide a custom error message describing the problem as well as a way to update the package via `pip`.

### DIFF
--- a/colcon_core/package_identification/python.py
+++ b/colcon_core/package_identification/python.py
@@ -7,7 +7,20 @@ from colcon_core.package_identification import logger
 from colcon_core.package_identification \
     import PackageIdentificationExtensionPoint
 from colcon_core.plugin_system import satisfies_version
-from setuptools.config import read_configuration
+try:
+    from setuptools.config import read_configuration
+except ImportError as e:
+    from pkg_resources import get_distribution
+    from pkg_resources import parse_version
+    setuptools_version = get_distribution('setuptools').version
+    minimum_version = '30.3.0'
+    if parse_version(setuptools_version) < parse_version(minimum_version):
+        e.msg += ', ' \
+            "'setuptools' needs to be at least version {minimum_version}, if" \
+            ' a newer version is not available from the package manager use ' \
+            "'pip3 install -U setuptools' to update to the latest version" \
+            .format_map(locals())
+    raise
 
 
 class PythonPackageIdentification(PackageIdentificationExtensionPoint):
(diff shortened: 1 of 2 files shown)


colcon_core/package_identification/python.py
```diff
@@ -2,12 +2,26 @@
...
-from setuptools.config import read_configuration
+try:
+    from setuptools.config import read_configuration
+except ImportError:
+    from pkg_resources import get_distribution
+    from pkg_resources import parse_version
+    setuptools_version = get_distribution('setuptools').version
+    minimum_version = '30.3.0'
+    if parse_version(setuptools_version) < parse_version(minimum_version):
+        sys.exit(
```
sloretz: `sys.exit()` is kind of harsh for an `ImportError`, but it does make the message easy to read when trying to bootstrap without installing setuptools.

```
$ ./src/colcon-core/bin/colcon
'setuptools' needs to be at least version 30.3.0, if a newer version is not available from the package manager use 'pip3 install -U setuptools' to update to the latest version
```

 I'm ok with it, but would prefer a raised `ImportError` instead.

dirk-thomas: I modified the patch to extend the raised `ModuleNotFound` exception with additional information where applicable: https://github.com/colcon/colcon-core/pull/91/commits/444e88edb9b2245d54388b37b751a1f821a6959a#diff-9fc4dc6d196841f3b396afba788b9246R18

The "downside" is that the build will continue, likely falling back to the `python_setup_py` extension which fails for every package using a `setup.cfg` file rather than declaring the package name in the `setup.py` file.

dirk-thomas: `ModuleNotFoundError` is actually only available as of Pytohn 3.6. Therefore I updated the patch to catch the more generic exception `ImportError`.

PR #93, opened 2018-08-02
add function to format duration as hours, minutes and seconds

dirk-thomas:
Instead of just printing the seconds (with two decimals) for it with minutes and potentially hours. Also use two, one or no decimals for the seconds depending on how large the duration is.

### DIFF
--- a/colcon_core/event_handler/__init__.py
+++ b/colcon_core/event_handler/__init__.py
@@ -103,3 +103,49 @@ def apply_event_handler_arguments(extensions, args):
         key = arg[:-1]
         extension = extensions[key]
         extension.enabled = suffix == '+'
+
+
+def format_duration(seconds, *, fixed_decimal_points=None):
+    """
+    Stringify a duration as hours, minutes, seconds and decimal points.
+
+    :param float seconds: The duration in seconds
+    :param int fixed_decimal_points: A fixed number of decimal points for the
+      seconds, if None two, one or none will be used depending on the duration
+      size
+    :returns: The string representation of the duration
+    :rtype: str
+    """
+    minutes, seconds = divmod(seconds, 60)
+    hours, minutes = divmod(minutes, 60)
+
+    # determine number of decimal points for seconds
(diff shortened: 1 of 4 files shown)


colcon_core/event_handler/__init__.py
```diff
@@ -103,3 +103,49 @@ def apply_event_handler_arguments(extensions, args):
         key = arg[:-1]
         extension = extensions[key]
         extension.enabled = suffix == '+'
+
+
+def format_duration(seconds, *, fixed_decimal_points=None):
+    """
+    Stringify a duration as hours, minutes, seconds and decimal points.
+
+    :param float seconds: The duration in seconds
```
sloretz: Maybe assert seconds > 0? It has some pretty weird results when negative

```
>>> print(format_duration(-1))
-1h 59min 59s
>>> print(format_duration(-3600))
-1h 0min 0s
>>> print(format_duration(-3601))
-2h 59min 59s
```

dirk-thomas: Good point. Added in e2ad19aa7325c408872569676760beb52518db9e.

colcon_core/event_handler/__init__.py
```diff
@@ -116,6 +116,9 @@ def format_duration(seconds, *, fixed_decimal_points=None):
     :returns: The string representation of the duration
     :rtype: str
     """
+    assert seconds >= 0.0, \
```
sloretz: Ah oops, my bad I shouldn't have said assert. This should raise `ValueError` so it doesn't get stripped if someone has `PYTHONOPTIMIZE` set.

dirk-thomas: Another good point :wink: b2f2d492f6bc9d2a2951214667f364522447b6de

dirk-thomas: And cover the new condition and line: 79432ce7bb3f10767b4c6221f0a550687723712e

Issue #106, opened 2018-09-04
Convert PackageDescriptor.dependencies into an object so metadata can be attached.


murphm8:
This work needs to be done to support colcon/colcon-ros#39.

Based on my initial investigation the following packages are affected:

* colcon-core #116
* ~~colcon-package-information~~
* ~~colcon-package-selection~~
* ~~colcon-parallel-executor~~
* ~~colcon-cargo~~
* ~~colcon-bazel~~
* ~~colcon-ros-bazel~~
* ~~colcon-cmake~~
* colcon-ros colcon/colcon-ros#40
* colcon-python-setup-py colcon/colcon-python-setup-py#11
* ~~colcon-gradle~~
* ~~colcon-ros-grade~~
* ~~colcon-powershell~~
* ~~colcon-graphviz-anim~~

I will start with colcon-core so we can settle on the API and usage and then follow up with PRs for all affected packages.

dirk-thomas:
I am wondering if it is necessary / useful to update all the packages which don't have any metadata or if we should just close those and let them keep adding plain `str` dependency names?

murphm8:
I think they can be closed for now, and if at some point those packages want to add metadata then that can be done in a PR with the metadata addition. 

Do you think the guidance from now on should be to use DependencyDescriptor?
Should consumers of dependencies that need the metadata just always verify the dependency is a DependencyDescriptor?

dirk-thomas:
We recently changed our bootstrap repo which contains all the Python packages so I am not 100% confident to do it now (with others not available) if something goes wrong. But I will aim to ship a new release of `colcon-core` on Monday.

dirk-thomas:
I am still blocked by problems with our bootstrap repo. I will comment here as soon as I was able to do a release... :disappointed:

dirk-thomas:
`colcon-core` 0.3.10 has just been released. Sorry for the delay. Please update the remaining PRs to include that version dependency - their CI should then pass and be mergable.

murphm8:
Awesome, thanks for getting these changes shipped! Everything for this ticket is completed. Will close it.

PR #107, opened 2018-09-04
Change PackageDescriptor.dependencies to DependencyDescriptors

murphm8:
This is for #106. 

I left some TODO in the code with questions. I think I caught all the usages in colcon_core. Anything using TaskContext.dependencies should be correct because that object is created in build.py/test.py where I have updated the usage of PackageDescriptor.dependencies.

I tested this with my updated colcon-ros in a simple ROS workspace. Do you have some test workspaces I can run this on to verify correctness? I don't have a python workspace to test against.

### DIFF
--- a/colcon_core/dependency_descriptor.py
+++ b/colcon_core/dependency_descriptor.py
@@ -1,0 +1,37 @@
+# Copyright 2016-2018 Dirk Thomas
+# Licensed under the Apache License, Version 2.0
+
+
+class DependencyDescriptor:
+    """
+    A descriptor for a package dependency.
+
+    A dependency is identified by its name.
+
+    The 'metadata' dictionary can store any additional information.
+    """
+
+    __slots__ = (
+        '_name',
+        'metadata'
+    )
+
+    def __init__(self, name, *, metadata=None):
+        """
+        Descriptor for a package dependency.
+
+        :param str name: Name of the declared dependency
+        :param dict metadata: Any metadata associated with
+        the dependency
+        """
+        self._name = name
+        self.metadata = metadata if metadata is not None else {}
+
+    @property
+    def name(self):
+        """Name of the dependency."""
+        return self._name
+
+    def __str__(self):  # noqa: D105
+        return '{' + ', '.join(
+            ['%s: %s' % (s, getattr(self, s)) for s in self.__slots__]) + '}'
--- a/colcon_core/dependency_descriptor.py
+++ b/colcon_core/dependency_descriptor.py
@@ -2,11 +2,12 @@
 # Licensed under the Apache License, Version 2.0
 
 
-class DependencyDescriptor:
+class DependencyDescriptor(str):
     """
     A descriptor for a package dependency.
 
-    A dependency is identified by its name.
+    A dependency is identified by its name. This subclasses str for backwards
+    compatibility purposes.
 
     The 'metadata' dictionary can store any additional information.
     """
@@ -16,14 +17,20 @@ class DependencyDescriptor:
         'metadata'
     )
 
-    def __init__(self, name, *, metadata=None):
+    def __new__(cls, name, metadata=None):
         """
         Descriptor for a package dependency.
 
-        :param str name: Name of the declared dependency
+        :param str value: Name of the declared dependency
         :param dict metadata: Any metadata associated with
         the dependency
         """
+        name_copy = str(name)
+        obj = str.__new__(cls, name_copy)
+        return obj
+
+    def __init__(self, name, metadata=None):  # noqa: D107
+        super(DependencyDescriptor, self).__init__()
         self._name = name
         self.metadata = metadata if metadata is not None else {}
 
@@ -32,9 +39,15 @@ class DependencyDescriptor:
         """Name of the dependency."""
         return self._name
 
-    def __str__(self):  # noqa: D105
-        return '{' + ', '.join(
-            ['%s: %s' % (s, getattr(self, s)) for s in self.__slots__]) + '}'
+    def __eq__(self, other):  # noqa: D105
+        if isinstance(other, DependencyDescriptor):
+            return self._name == other._name and \
+                self.metadata == other.metadata
+        else:
+            return self._name == other
+
+    def __hash__(self):  # noqa: D105
+        return hash(self._name)
 
 
 def dependency_name(dependency):
--- a/colcon_core/dependency_descriptor.py
+++ b/colcon_core/dependency_descriptor.py
@@ -35,3 +35,24 @@ class DependencyDescriptor:
     def __str__(self):  # noqa: D105
         return '{' + ', '.join(
             ['%s: %s' % (s, getattr(self, s)) for s in self.__slots__]) + '}'
+
+
+def dependency_name(dependency):
+    """
+    Get the name of the dependency.
+
+    This should be used to maintain backwards compatibility with extensions
+    that treat PackageDescriptor.dependencies as a list of strings
+    instead of a list of DependencyDescriptor
+
+    :param dependency: string or DependencyDescriptor
+    :return: name of the dependency
+    :rtype: str
+    """
+    if isinstance(dependency, DependencyDescriptor):
+        return dependency.name
+    elif isinstance(dependency, str):
+        return dependency
+    else:
+        raise RuntimeError(
+            'Passed in object is not a str or DependencyDescriptor')
--- a/colcon_core/environment/__init__.py
+++ b/colcon_core/environment/__init__.py
@@ -6,6 +6,7 @@ import os
 from pathlib import Path
 import traceback
 
+from colcon_core.dependency_descriptor import dependency_name
 from colcon_core.location import get_relative_package_index_path
 from colcon_core.logging import colcon_logger
 from colcon_core.plugin_system import instantiate_extensions
@@ -126,7 +127,8 @@ def create_environment_scripts(
     # create file containing the runtime dependencies
     path = prefix_path / get_relative_package_index_path() / pkg.name
     path.parent.mkdir(parents=True, exist_ok=True)
-    dependencies = [d.name for d in pkg.dependencies.get('run', set())]
+    dependencies = [
+        dependency_name(d) for d in pkg.dependencies.get('run', set())]
     path.write_text(os.pathsep.join(sorted(dependencies)))
 
 
--- a/colcon_core/environment/__init__.py
+++ b/colcon_core/environment/__init__.py
@@ -126,8 +126,8 @@ def create_environment_scripts(
     # create file containing the runtime dependencies
     path = prefix_path / get_relative_package_index_path() / pkg.name
     path.parent.mkdir(parents=True, exist_ok=True)
-    path.write_text(
-        os.pathsep.join(sorted(pkg.dependencies.get('run', set()))))
+    dependencies = [d.name for d in pkg.dependencies.get('run', set())]
+    path.write_text(os.pathsep.join(sorted(dependencies)))
 
 
 def create_environment_hooks(prefix_path, pkg_name):
--- a/colcon_core/package_descriptor.py
+++ b/colcon_core/package_descriptor.py
@@ -14,7 +14,7 @@ class PackageDescriptor:
     * the 'type' which must be a non-empty string
     * the 'name' which must be a non-empty string
 
-    The names of 'dependencies' are grouped by their category.
+    'dependencies' are grouped by their category as DependencyDescriptor
 
     Each item in 'hooks' must be a relative path in the installation space.
 
@@ -58,8 +58,8 @@ class PackageDescriptor:
         Get the dependencies for specific categories or for all categories.
 
         :param Iterable[str] categories: The names of the specific categories
-        :returns: The names of the dependencies
-        :rtype: set
+        :returns: The dependencies for the passed in category or all
+        :rtype: set[DependencyDescriptor]
         :raises AssertionError: if the package name is listed as a dependency
         """
         dependencies = set()
@@ -67,7 +67,8 @@ class PackageDescriptor:
             categories = self.dependencies.keys()
         for category in sorted(categories):
             dependencies |= self.dependencies[category]
-        assert self.name not in dependencies, \
+        dependency_names = {d.name for d in dependencies}
+        assert self.name not in dependency_names, \
             "The package '{self.name}' has a dependency with the same name" \
             .format_map(locals())
         return dependencies
@@ -87,17 +88,18 @@ class PackageDescriptor:
           categories
         :param Iterable[str] recursive_categories: The names of the recursive
           categories
-        :returns: The names of the recursive dependencies
-        :rtype: set
+        :returns: The recursive dependencies of self
+        :rtype: set[DependencyDescriptor]
         :raises AssertionError: if a package lists itself as a dependency
         """
         queue = self.get_dependencies(categories=direct_categories)
         dependencies = set()
         while queue:
             # pick one dependency from the queue
-            name = queue.pop()
+            dependency = queue.pop()
+            name = dependency.name
             # ignore redundant dependencies
-            if name in dependencies:
+            if name in {dep.name for dep in dependencies}:
                 continue
             # ignore circular dependencies
             if name == self.name:
@@ -112,7 +114,7 @@ class PackageDescriptor:
             for d in descs:
                 queue |= d.get_dependencies(categories=recursive_categories)
             # add dependency to result set
-            dependencies.add(name)
+            dependencies.add(dependency)
         return dependencies
 
     def __hash__(self):  # noqa: D105
--- a/colcon_core/package_descriptor.py
+++ b/colcon_core/package_descriptor.py
@@ -16,7 +16,7 @@ class PackageDescriptor:
     * the 'type' which must be a non-empty string
     * the 'name' which must be a non-empty string
 
-    'dependencies' are grouped by their category as DependencyDescriptor
+    'dependencies' are grouped by their category as DependencyDescriptor or str
 
     Each item in 'hooks' must be a relative path in the installation space.
 
@@ -61,7 +61,7 @@ class PackageDescriptor:
 
         :param Iterable[str] categories: The names of the specific categories
         :returns: The dependencies for the passed in category or all
-        :rtype: set[DependencyDescriptor]
+        :rtype: set[DependencyDescriptor, str]
         :raises AssertionError: if the package name is listed as a dependency
         """
         dependencies = set()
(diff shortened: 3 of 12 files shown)


colcon_core/command.py
```diff
@@ -63,7 +63,7 @@ def main(*, command_name='colcon', argv=None):
     """
     global colcon_logger
     # default log level
-    colcon_logger.setLevel(logging.WARN)
+    colcon_logger.setLevel(logging.INFO)
```
dirk-thomas: Please revert this unrelated change.

colcon_core/dependency_descriptor.py
```diff
@@ -0,0 +1,36 @@
+# Copyright 2016-2018 Dirk Thomas
+# Licensed under the Apache License, Version 2.0
+
+
+class DependencyDescriptor:
+    """A descriptor for a package dependency."""
+
```
dirk-thomas: Please use `__slots__` to limit the members to a known set (`name`, `metadata`).

colcon_core/dependency_descriptor.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init__(self, name, metadata=None):
```
dirk-thomas: Force the keyword argument: `def __init__(self, name, *, metadata=None):`

colcon_core/dependency_descriptor.py
```diff
@@ -0,0 +1,36 @@
...
+        :param str name: Name of the declared dependency
+        :param dict metadata: Any metadata associated with
+        the dependency
+        """
+        self.name = name
+
+        if metadata is None:
+            self.metadata = {}
+        else:
+            self.metadata = metadata
```
dirk-thomas: Nipick: shorten to `self.metadata = metadata if metadata is not None else {}`.

colcon_core/dependency_descriptor.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def __eq__(self, other):  # noqa: D105
+        return self.name == other.name
+
+    def __lt__(self, other):  # noqa: D105
+        return self.name < other.name
+
+    def __hash__(self):  # noqa: D105
+        return hash(self.name)
```
dirk-thomas: Generally this is dangerous since after an instance has been hashed the name could be changed which would result in a different hash. I don't have a better proposal though. Maybe prevent setting the `name` outside of the constructor in this case?

colcon_core/dependency_descriptor.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def __str__(self):  # noqa: D105
+        return self.name
+
+    def __repr__(self):  # noqa: D105
+        return 'DependencyDescriptor({name})'.format(name=self.name)
```
dirk-thomas: I would rather not define `__repr__` is not necessarily for anything?

colcon_core/environment/__init__.py
```diff
@@ -126,8 +126,12 @@ def create_environment_scripts(
     # create file containing the runtime dependencies
     path = prefix_path / get_relative_package_index_path() / pkg.name
     path.parent.mkdir(parents=True, exist_ok=True)
+    dependencies = [dependency.name
+                    for dependency in pkg.dependencies.get('run', set())]
```
dirk-thomas: Nitpick: shorten local variable name to: `dependencies = [d.name for d in pkg.dependencies.get('run', set())]`

In general I try to avoid aligned indentation. Maybe the new code fits in a single line. If not please wrap after the open parenthesis / bracket.

colcon_core/environment/__init__.py
```diff
@@ -126,8 +126,12 @@ def create_environment_scripts(
...
+    # TODO: Is this supposed to be in topological order? I don't think the
+    # set is ordered?
```
dirk-thomas: No, these dependencies are not ordered. Alphabetical only for deterministic results.

colcon_core/package_descriptor.py
```diff
@@ -67,7 +67,8 @@ def get_dependencies(self, *, categories=None):
             categories = self.dependencies.keys()
         for category in sorted(categories):
             dependencies |= self.dependencies[category]
-        assert self.name not in dependencies, \
+        dependency_names = {dependency.name for dependency in dependencies}
```
dirk-thomas: Same: `dependency_names = {d.name for d in dependencies}

colcon_core/package_identification/python.py
```diff
@@ -100,5 +101,5 @@ def extract_dependencies(options):
             # remove environmental markers (separated by semicolons)
             # and version specifiers (separated by comparison operators)
             name = re.split(r';|<|>|<=|>=|==|!=', dep)[0].rstrip()
-            dependencies[dependency_type].add(name)
+            dependencies[dependency_type].add(DependencyDescriptor(name))
```
dirk-thomas: The version constraints of the string should be extracted and added to the descriptor. This could be done in a separate patch if you prefer though.

colcon_core/topological_order.py
```diff
@@ -39,14 +39,18 @@ def topological_order_packages(
         # remove dependencies on already ordered packages
         ordered_names = {descriptor.name for descriptor in ordered.keys()}
         for q in queued:
-            q.remaining_dependencies -= ordered_names
+            deps = set(filter(lambda dependency:
+                              dependency.name not in ordered_names,
+                              q.remaining_dependencies))
+            q.remaining_dependencies = deps
```
dirk-thomas: Instead of modifying the algorithm (which doesn't need to know about versions) I would prefer changing the input to still only contain the names rather than full dependency descriptors (I think `rec_deps` above).

colcon_core/verb/test.py
```diff
@@ -214,7 +214,9 @@ def _get_jobs(self, args, decorators, install_base):
 
             recursive_dependencies = OrderedDict()
             # for testing a package include itself in the environment
-            for dep_name in decorator.recursive_dependencies + [pkg.name]:
+            dependency_names = [dep.name
```
dirk-thomas: Nitpick: please wrap after the bracket.

test/test_package_augmentation.py
```diff
@@ -107,29 +108,34 @@ def test_update_descriptor():
     assert len(desc.metadata) == 0
 
     data = {
-        'build-dependencies': {'b1', 'b2'},
-        'test-dependencies': {'t1'},
+        'build-dependencies': {DependencyDescriptor('b1'),
```
dirk-thomas: Nitpick: please wrap after the curly brace.

Same below.

test/test_package_descriptor.py
```diff
@@ -33,15 +34,18 @@ def test_identifies_package():
...
-    d1.dependencies['run'].add('run-depend')
-    d1.dependencies['run'].add('depend')
-    assert d1.get_dependencies() == {'build-depend', 'run-depend', 'depend'}
-
-    d1.dependencies['test'].add('self')
+    d1.dependencies['build'].add(DependencyDescriptor('build-depend'))
+    d1.dependencies['build'].add(DependencyDescriptor('depend'))
+    d1.dependencies['run'].add(DependencyDescriptor('run-depend'))
+    d1.dependencies['run'].add(DependencyDescriptor('depend'))
+    # TODO: What if the run and build dependencies have different versions?
```
dirk-thomas: I would say the more strict version constraint should be used when merging. In case of a conflict an `error` message should be logged.

murphm8: Should this logic be on the implementer of the package identification extension point instead?
 
We could then say that in colcon-core there will be no conflicting dependencies.

colcon_core/environment/__init__.py
dirk-thomas: I think the comment it still valid as it is. The file contains "the runtime dependencies" and it is being used to "source all packages in topological order", It doesn't imply that the order in the file itself is ordered.

colcon_core/dependency_descriptor.py
murphm8: Converted name to be a read-only property. 

The other issue I see here is that two DependencyDescriptor with different metadata but the same name will be considered equal. Should __eq__ contain the metadata dictionary too? That would require everything stored in metadata to be comparable, is that okay?

colcon_core/environment/__init__.py
```diff
@@ -126,8 +126,12 @@ def create_environment_scripts(
...
     path.write_text(
-        os.pathsep.join(sorted(pkg.dependencies.get('run', set()))))
+        os.pathsep.join(sorted(dependencies)))
```
murphm8: This functions docstring states:     

:returns: iterable of the generated hook paths
:rtype: Iterable

But there is no return statement, is that an issue?

colcon_core/package_identification/python.py
murphm8: Yea, I'll do that in a separate commit.

colcon_core/dependency_descriptor.py
dirk-thomas: Another option would be to just not override `__eq__` and wait until there is a use case for it.

colcon_core/environment/__init__.py
dirk-thomas: I think the docstring is just wrong. Please address this in a separate PR (otherwise I can do that too if you prefer).

colcon_core/topological_order.py
murphm8: That's much simpler! Could you explain the difference between PackageDecorator and PackageDescriptor? I'm concerned PackageDecorator.recursive_dependencies is a different type than PackageDescriptor.dependencies if I simplify this.

dirk-thomas: https://github.com/colcon/colcon-core/blob/bdba48223fbd73448caf424a934faaec35e8b4ee/colcon_core/package_descriptor.py#L8-L22

https://github.com/colcon/colcon-core/blob/bdba48223fbd73448caf424a934faaec35e8b4ee/colcon_core/package_decorator.py#L16-L24

murphm8:
Hey, I'm going to be on vacation the rest of the week but will pick this up on Monday. Thanks for all your feedback!

colcon_core/dependency_descriptor.py
murphm8: Overriding `__eq__` is nice for unit testing, not sure if that is enough of a reason? 

Should I just make `__eq__` use `(self.name, self.metadata)`, and have `__hash__` just use `self.name`?

murphm8: It's needed in a couple different tests where we check dependencies. It would also be needed by any consumers of colcon-core that want to test using DependencyDescriptor.

I think keeping `__eq__` would be simpler.

dirk-thomas: > It would also be needed by any consumers of colcon-core that want to test using DependencyDescriptor.

Do we know what potential consumers want to test for? If not, I would rather leave it out for now until that is the case. We can always add it later when there is a concrete need for it.

murphm8: Looking into what happens when I remove `__eq__`, we get duplicates in the set. If I remove `__eq__` then I see unit tests where the same dependency is listed twice in the set.

    def test_get_dependencies():
        d1 = PackageDescriptor('/some/path')
        d1.name = 'self'
        d1.dependencies['build'].add(DependencyDescriptor('build-depend'))
        d1.dependencies['build'].add(DependencyDescriptor('depend'))
        d1.dependencies['run'].add(DependencyDescriptor('run-depend'))
        d1.dependencies['run'].add(DependencyDescriptor('depend'))
    >       assert check_dependencies(d1.get_dependencies(),
                                  ['build-depend', 'run-depend', 'depend'])
    E       AssertionError: assert False
    E        +  where False = check_dependencies({{_name: build-depend, metadata: {}}, {_name:     depend, metadata: {}}, {_name: depend, metadata: {}}, {_name: run-depend, metadata: {}}},     ['build-depend', 'run-depend', 'depend'])
    E        +    where {{_name: build-depend, metadata: {}}, {_name: depend, metadata: {}},     {_name: depend, metadata: {}}, {_name: run-depend, metadata: {}}} = <bound method PackageDescriptor.get_dependencies of <colcon_core.package_descriptor.PackageDescriptor object at 0x1074a3db0>>()
    E        +      where <bound method PackageDescriptor.get_dependencies of <colcon_core.package_descriptor.PackageDescriptor object at 0x1074a3db0>> = <colcon_core.package_descriptor.PackageDescriptor object at 0x1074a3db0>.get_dependencies

Above we can see that `d1.get_dependencies()` returns a set containing the "depend" dependency twice. If I define `__eq__` as `(this.name, this.metadata) == (other.name, other.metadata)` then there are no duplicates. 

It's my understanding that an object with a matching hash is not necessarily equal. So `obj1.hash == obj2.hash` doesn't imply `obj1 == obj2`, but `obj1 == obj2` implies `obj1.hash == obj2.hash`. This means that the set semantics will keep both copies of the object if `__eq__` is not defined.

dirk-thomas: > but `obj1 == obj2` implies `obj1.hash == obj2.hash`.

Indeed, looking at the Python documentation of [__hash__](https://docs.python.org/3/reference/datamodel.html#object.__hash__):

    "The only required property is that objects which compare equal have the same hash value"

> Above we can see that d1.get_dependencies() returns a set containing the "depend" dependency twice.

The return value of `get_dependencies()` *should* contain "duplicates" because two dependencies with the same name do not necessarily have the same version constraints.

So if you store dependencies in a `set` (as this function does in the return value) it *must not* de-duplicate dependencies solely based on their name since the metadata might be different. As a direct consequence the hash can't be computed based on the name alone.

The question if dependencies which have the same name and metadata should be de-duplicated is a separate one: the function itself could (not must) de-duplicate dependencies which are equal in all the properties. That could either happen within the logic of the function or if `__hash__` is defined across all properties (instead of relying on the object id). And again in this case my previous concern applies again: exposing the metadata which allows callers to manipulate the data *after* a dependency has been hashed is dangerous.

Therefore I am still advocating to neither define `__hash__` nor `__eq__` since I don't see the need for them.

colcon_core/dependency_descriptor.py
```diff
@@ -0,0 +1,40 @@
...
+        self._name = name
+        self.metadata = metadata if metadata is not None else {}
+
+    @property
+    def name(self):
+        """Name of the dependency."""
+        return self._name
+
+    def __lt__(self, other):  # noqa: D105
+        return self.name < other.name
```
dirk-thomas: Why is this one necessary?

colcon_core/package_descriptor.py
```diff
@@ -87,17 +88,18 @@ def get_recursive_dependencies(
           categories
         :param Iterable[str] recursive_categories: The names of the recursive
           categories
-        :returns: The names of the recursive dependencies
+        :returns: DependencyDescriptor objects of recursive dependencies
```
dirk-thomas: Please mention the type of the return value as part of the `rtype`:

```
:returns: The recursive dependencies
:rtype: set[DependencyDescriptor]
```

colcon_core/dependency_descriptor.py
murphm8: Removed

colcon_core/package_descriptor.py
murphm8: Updated both docstrings.

murphm8:
To roll this out do we need to make a pre-change to all the packages that would break? Something like lock their colcon-core dependency to <= 0.3.9. Then make the change and update them all to depend on >= 0.3.10?

dirk-thomas:
> Something like lock their colcon-core dependency

That wouldn't help since we can't guarantee that user will actually get that interim version. They might just stick with the version the already have.

murphm8:
How have you handled breaking changes like this in the past? Is there anything I can do?

dirk-thomas:
> How have you handled breaking changes like this in the past?

Since the versions are still 0.x it is technically fine to break anything in a minor release. That being said it is obviously not desired when it can be avoided. Also looking into the future it would be good to figure out the right approach how to handle this (if we would be post 1.0 and don't want to level the major version).

> Is there anything I can do?

I have to think about it (when I find a bit more time) if there is a way to do it without breakage (e.g. subclassing from `str`) or what should be changed so that this is catchable gracefully in the future.

murphm8:
I pushed a new commit that adds backwards compatibility with existing extensions that treat PackageDescriptor.dependencies as a list of strings. I've  updated the tests to include both DependencyDescriptor and str dependencies in the dependencies list.

dirk-thomas:
I don't see how the updated patch achieves backward compatibility. Existing code is expecting a `str` and doesn't know about / use the new function `dependency_name`. I haven't tried the patch but don't see how that would work in a backward compatible fashion. Can you please elaborate how you tested backward compatibility (e.g. using this branch of `colcon-core` in combination with existing released extensions).

murphm8:
You're correct. This will break existing consumers that depend on PackageDescriptor.dependencies being a list of strings. 

The only way I can think to support both would be to create a new attribute on PackageDescriptor for DependencyDescriptors. Then a setter and getter for PackageDescriptor.dependencies. 

The setter would take a list of strings and create the PackageDescriptor.dependeny_descriptors list. 

The getter would return [d.name for d in descriptor.dependency_descriptors]

What do you think of that approach? Would you be okay with a new attribute on PackageDescriptor?

murphm8:
I'm looking into subclassing str. The one issue I'd have with that is then anywhere that wants to access metadata on a DependencyDescriptor needs to first verify isInstance(d, DependencyDescriptor). Do you think that's okay?

murphm8:
With this change I've verified that the colcon-package-information (colcon info) in PyPI works with the latest colcon-core and colcon-ros.

dirk-thomas:
I reviewed the current state of the patch and a few things looked "weird" to me - or at least they were unclear why they are necessary:

* inheriting from `str` but still having a `_name` slot
* inheriting from `str` but still having `__slots__`
* (the need to) override the `__eq__` / `__hash__` methods
* the need for `dependency_name` - the goal is that existing code not using that function are still compatible
* the "logic" changes in `get_dependencies` / `get_recursive_dependencies` to use the name rather then the descriptor

Instead of iterating on this patch I tried locally if we can't get away with a much smaller patch. I prototyped #116 and it seems to work for me. Can you please give that patch a try and comment if it works for you too.

murphm8:
I agree your change is much cleaner and more concise. I found I needed to override __eq__ and __hash__ to get DependencyDescriptor to work correctly as a `str` subclass. 

The `dependency_name` function could go away. The changes in `get_dependencies` and `get_recursive_dependencies` would also not be necessary if DependencyDescriptor can be treated as a string.

dirk-thomas:
> The changes in get_dependencies and get_recursive_dependencies would also not be necessary if DependencyDescriptor can be treated as a string.

I included these changes so that "consumers" of the functions get a consistent return value and don't have to deal with two potential types.

dirk-thomas:
> I found I needed to override __eq__ and __hash__ to get DependencyDescriptor to work correctly as a str subclass.

I didn't see a need for it. Can you double check if the current state works for you (without these methods overridden)?

murphm8:
> I didn't see a need for it. Can you double check if the current state works for you (without these methods overridden)?

Posted on #116. Your change works for me.

> I included these changes so that "consumers" of the functions get a consistent return value and don't have to deal with two potential types.

I agree doing this will make it easier for consumers of those apis.

dirk-thomas:
Closing in favor of #116.

PR #116, opened 2018-10-08
add descriptor for package dependencies

dirk-thomas:
Replaces #107.

### DIFF
--- a/colcon_core/package_descriptor.py
+++ b/colcon_core/package_descriptor.py
@@ -4,6 +4,8 @@
 from collections import defaultdict
 from pathlib import Path
 
+from colcon_core.dependency_descriptor import DependencyDescriptor
+
 
 class PackageDescriptor:
     """
@@ -14,7 +16,8 @@ class PackageDescriptor:
     * the 'type' which must be a non-empty string
     * the 'name' which must be a non-empty string
 
-    The names of 'dependencies' are grouped by their category.
+    'dependencies' are grouped by their category as `DependencyDescriptor` or
+    `str`.
 
     Each item in 'hooks' must be a relative path in the installation space.
 
@@ -58,8 +61,8 @@ class PackageDescriptor:
         Get the dependencies for specific categories or for all categories.
 
         :param Iterable[str] categories: The names of the specific categories
-        :returns: The names of the dependencies
-        :rtype: set
+        :returns: The dependencies
+        :rtype: set[DependencyDescriptor]
         :raises AssertionError: if the package name is listed as a dependency
         """
         dependencies = set()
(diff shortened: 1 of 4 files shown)


murphm8:
I'm getting this error when attempting to build a python package with a setup.py (I've modified `colcon-python-setup-py`):

    File "/workspace/src/colcon-core/colcon_core/environment/__init__.py", line 130, in create_environment_scripts
    os.pathsep.join(sorted(pkg.dependencies.get('run', set()))))
    TypeError: sequence item 0: expected str instance, DependencyDescriptor found

It looks like os.pathsep.join requires `str` and not `collections.UserString(Sequence)`.

murphm8:
Changing that line to:

    os.pathsep.join(sorted(map(lambda s: str(s), pkg.dependencies.get('run', set()))))

Fixes the bug, but it looks like `collections.UserString` is not backwards compatible.

dirk-thomas:
Hm, I would have expected `collections.UserString` to behave nicer :disappointed: It sounds like sticking to `str` as the base class as in your PR is the "more" backward compatible choice. I updated the PR to do so in 998056c261d4ecd333713aebc554036f79c39bc3.

murphm8:
Ran your change against my test workspace and it worked correctly. Ship it!

Once this is in master I will make a pull request for the python version metadata change. Do you also want a pull request with the changes for get_dependencies and get_recursive_dependencies?

dirk-thomas:
> Once this is in master I will make a pull request for the python version metadata change.

I guess you mean this patch: https://github.com/murphm8/colcon-core/commit/9a6e4a8e9f49a9d103cfb37712536e88a7203fbf ?

> Do you also want a pull request with the changes for get_dependencies and get_recursive_dependencies?

Sorry, if I missed this but can you clarify what exact change you are referring to here?

> what happens when we have two DependencyDescriptors with different metadata but the same name? In my PR they would not be equal, but in this PR will they be equal because as `str` it will just compare the value of the string?

By inheriting from `str` the behavior is unchanged. A dependency is uniquely identified by its name (https://github.com/colcon/colcon-core/pull/116/files#diff-855ecc0b44feadfc4370ae876811325dR12). If e.g. a package identification extension has two version constraints for the same dependency it needs to merge them into a single dependency descriptor.

dirk-thomas:
> I don't see those changes here so ...

This change: https://github.com/colcon/colcon-core/pull/116/files#diff-3ef2dc6c9888dd64480ca0888e4849c0R76 ?

murphm8:
> This change: https://github.com/colcon/colcon-core/pull/116/files#diff-3ef2dc6c9888dd64480ca0888e4849c0R76 ?

Ah, I missed it, that looks good.

colcon_core/package_descriptor.py
```diff
@@ -70,25 +73,27 @@ def get_dependencies(self, *, categories=None):
         assert self.name not in dependencies, \
             "The package '{self.name}' has a dependency with the same name" \
             .format_map(locals())
-        return dependencies
+        return {
+            (DependencyDescriptor(d) if d is str else d)
```
dirk-thomas: The condition `d is str` is always `False`. See #173 for the fix.

PR #117, opened 2018-10-10
remove unused variable

dirk-thomas:
Removing `e` since it isn't being used in the exception handling block.


dirk-thomas:
a525247ec3804350be7e3959c055a7d9c2b7b5a1 removes a few more unused variables.

PR #120, opened 2018-10-12
Update python incompatible version algorithm to use tuple unpacking

murphm8:
I just read a blog [post](http://treyhunner.com/2018/10/asterisks-in-python-what-they-are-and-how-to-use-them/) and was inspired to create a more pythonic solution for this algorithm, thought I'd share.


dirk-thomas:
LGTM. Thanks for the improvement.

dirk-thomas:
For the record this is a follow up of #118.

PR #122, opened 2018-10-21
Bug when setting a python executable and it is not used

Bimble:
Came across this problem when trying to use a python executable but it was not detected. it was comparing an empty string.

A set is done inside the if statement and evaluated inside the if statement, therefore the variable should use delayed expansion as variable expansion occurs when the if statement is evaluated. 

More info about the topic can be found [here](https://blogs.msdn.microsoft.com/oldnewthing/20060823-00/?p=29993).


dirk-thomas:
Thank you for the patch.

PR #132, opened 2018-11-21
Enabling symlink creation on Windows

brawner:
This adds symlink creation for Windows for 'latest {Verb}'. Currently os.symlink only works when run with administrator privileges and it will throw an OSError if symlink privileges are not available.

### DIFF
--- a/colcon_core/location.py
+++ b/colcon_core/location.py
@@ -3,7 +3,6 @@
 
 import os
 from pathlib import Path
-import sys
 import uuid
 
 from colcon_core.logging import colcon_logger
@@ -121,10 +120,12 @@ def create_log_path(verb_name):
     A `COLCON_IGNORE` marker file is being placed in the parent directory of
     the logging directory to avoid it being crawled for packages.
 
-    On non-Windows platforms two symlinks are created as siblings of the log
-    path:
+    Two symlinks are created as siblings of the log path:
     * `latest_<verb_name>` linking to the log path
     * `latest` linking to `latest_<verb_name>`
+
+    On Windows platforms, Administrator privilges are required to create
+    these symlinks.
 
     :param str verb_name: The verb name
     """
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Please address the linter warning and update the corresponding test to not skip Windows.

brawner:
Is it ok to assume the test will be run with admin privileges on Windows?

dirk-thomas:
I don't know what AppVeyor uses.

colcon_core/location.py
```diff
@@ -121,11 +120,13 @@ def create_log_path(verb_name):
     A `COLCON_IGNORE` marker file is being placed in the parent directory of
     the logging directory to avoid it being crawled for packages.
 
-    On non-Windows platforms two symlinks are created as siblings of the log
-    path:
+    Two symlinks are created as siblings of the log path:
     * `latest_<verb_name>` linking to the log path
     * `latest` linking to `latest_<verb_name>`
 
+    On Windows platforms, Administrator privilges are required to create
```
dirk-thomas: Spelling: "privilges"

brawner:
```suggestion
    On Windows platforms, Administrator privileges are required to create
```

brawner: Done

davetcoleman:
Great contribution!

brawner:
Thanks Dave!

dirk-thomas:
Can you please update the test to also mock the `symlink` function and raise the expected exception to cover these new lines.

dirk-thomas:
@brawner Friendly ping.

brawner:
@dirk-thomas In order to raise the exception, the user must be demoted to standard privileges and I cannot find a way to do this on Windows in the middle of a Python script. Changing permissions in the directory to read-only will raise a different exception (PermissionsError).

Are there any other alternatives I'm not thinking about?

dirk-thomas:
> Are there any other alternatives I'm not thinking about?

You can mock the Python function to raise the expected exception.

dirk-thomas:
Thank you for the patch.

Issue #134, opened 2018-12-03
Deprecation warnings: Using or importing the ABCs from 'collections'


brawner:
For the Python 3.7 users these warnings pop up in many places in ROS2. They'll need to be addressed in order for compatibility with Python 3.8 (currently to be released in late 2019).

```
__________________ ERROR collecting test/test_environment.py __________________
test\test_environment.py:8: in <module>
    from colcon_core.environment import create_environment_hooks
colcon_core\environment\__init__.py:4: in <module>
    from collections import Iterable
<frozen importlib._bootstrap>:1032: in _handle_fromlist
    ???
c:\python37\lib\collections\__init__.py:52: in __getattr__
    DeprecationWarning, stacklevel=2)
E   DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
_______________ ERROR collecting test/test_environment_path.py ________________
test\test_environment_path.py:7: in <module>
    from colcon_core.environment.path import PathEnvironment
colcon_core\environment\__init__.py:4: in <module>
    from collections import Iterable
<frozen importlib._bootstrap>:1032: in _handle_fromlist
    ???
c:\python37\lib\collections\__init__.py:52: in __getattr__
    DeprecationWarning, stacklevel=2)
E   DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
____________ ERROR collecting test/test_environment_pythonpath.py _____________
test\test_environment_pythonpath.py:8: in <module>
    from colcon_core.environment.pythonpath import PythonPathEnvironment
colcon_core\environment\__init__.py:4: in <module>
    from collections import Iterable
<frozen importlib._bootstrap>:1032: in _handle_fromlist
    ???
c:\python37\lib\collections\__init__.py:52: in __getattr__
    DeprecationWarning, stacklevel=2)
E   DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
- generated xml file: C:\Users\brawner\ros2_core\build\colcon-core\pytest.xml -
```

brawner:
I should also point out, these cause test failures when running tests with Python 3.7.

dirk-thomas:
> They'll need to be addressed in order for compatibility with Python 3.8

Please consider to provide a pull request for the proposed change.

> these cause test failures when running tests with Python 3.7.

[...]

PR #136, opened 2018-12-07
collections.abc

brawner:
This updates an import statement with regards to the collections deprecated namespace.

This addresses issue #136


dirk-thomas:
Thanks for the patch.

For the record: this addresses issue #134.

PR #144, opened 2019-01-02
add OnError enum to configure the executor behavior when a job fails

dirk-thomas:
Adds an `OnError` enum to offer more fine grain control about the executor behavior in case a job fails. The new keyword argument is introduced in a backward compatible way so that existing code calling the global function `colcon_core.executor.execute_jobs` as well as existing `ExecutorExtensionPoint.execute` methods continue to work.

The patch does neither modify the verbs to make use of the additional on-error values nor the executors to make use of the new keyword argument.

### DIFF
--- a/test/test_executor.py
+++ b/test/test_executor.py
@@ -15,6 +15,7 @@ from colcon_core.executor import execute_jobs
 from colcon_core.executor import ExecutorExtensionPoint
 from colcon_core.executor import get_executor_extensions
 from colcon_core.executor import Job
+from colcon_core.executor import OnError
 from colcon_core.subprocess import SIGINT_RESULT
 from mock import Mock
 from mock import patch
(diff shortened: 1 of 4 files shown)


test/test_executor.py
```diff
@@ -224,8 +225,9 @@ def test_execute_jobs():
             event_reactor.get_queue().put.reset_mock()
             jobs['one'].returncode = 0
             extensions = get_executor_extensions()
-            extensions[110]['extension2'].execute = Mock(return_value=0)
-            rc = execute_jobs(context, jobs)
+            extensions[110]['extension2'].execute = \
```
mjcarroll: I can't tell what the reason for `110` here is.  Is this just to ensure that you are above any other extensions that may have been in the list?

dirk-thomas: Yes, the extensions returned by [get_executor_extensions](https://github.com/colcon/colcon-core/blob/35f04c4289c26b163f14989fe23989f6e7440a83/colcon_core/executor/__init__.py#L169-L177) are first grouped and ordered by their priorities and then in a dictionary with their name as the key. Some extension points do have a priority (to influence their order of use) while others don't (when there is no need for an order). For this specific test the extension has the priority 110 (see https://github.com/colcon/colcon-core/blob/cc235f558756c882b3d6f0076cb448c0f427dd3a/test/test_executor.py#L136) and therefore is returned in this nested dictionary under that priority group.

PR #148, opened 2019-01-08
add get_previous_log_path

dirk-thomas:
Required for colcon/colcon-package-selection#17.


dirk-thomas:
After thinking about this approach a bit more I think it is too fragile. If the `log` directories are removed the state information is gone. If the build directory is being removed the state information is wrong.

Therefore I will close this in favor of a new approach storing the information in the build directory.

PR #163, opened 2019-02-08
fix the order of prepended PYTHONPATH in develop mode if other paths are set in PYTHONPATH

dirk-thomas:
Fixes ament/ament_cmake#152.

E.g. the environment variable `PYTHONPATH` contains already some paths - e.g. `/tmp` - and a package `foo` is being installed in `develop` mode. The resulting `PYTHONPATH` is `<install>/foo/<python-lib>:/tmp` which looks correct.

Though at runtime the `sys.path` is being extended when the `easy-install.pth` file in `<install>/foo/<python-lib>` is found and `<build>/foo` is being added - but *after* all paths listed in the `PYTHONPATH`. That implies that the resolution order changes by considering `/tmp` before `<build>/foo`.

This is in particular a problem if the package `foo` is supposed to be an *overlay* of the same package already present in a directory already mentioned in `PYTHONPATH`. Due to the changed order `foo` would be used from the underlay rather than the intended overlay.

### DIFF
--- a/colcon_core/task/python/build.py
+++ b/colcon_core/task/python/build.py
@@ -4,6 +4,7 @@
 import asyncio
 from distutils.sysconfig import get_python_lib
 import os
+from pathlib import Path
 import re
 import shutil
 import sys
@@ -13,6 +14,7 @@ from colcon_core.environment import create_environment_hooks
 from colcon_core.environment import create_environment_scripts
 from colcon_core.logging import colcon_logger
 from colcon_core.plugin_system import satisfies_version
+from colcon_core.shell import create_environment_hook
 from colcon_core.shell import get_command_environment
 from colcon_core.task import check_call
 from colcon_core.task import TaskExtensionPoint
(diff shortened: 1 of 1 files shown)


dirk-thomas:
> My first observation is that we are still using `COLCON_CURRENT_PREFIX` (which maps to `/path/to/workspace/install/package_name`) as prefix of the build path. Will this work if `build` folder is not next to `install`?

The path `$COLCON_CURRENT_PREFIX/../../build/ros2cli/share/ros2cli/hook/pythonpath_develop.sh` is computed and changes with the relative location of `build` from `install`. So I would expect this to work.

> the generated value is not correct: `COLCON_CURRENT_PREFIX` is not needed in this case; the path that we want to add would be just `/home/juan/ros2cli_ws/build/ros2cli`.

Are you sure you are trying this with the current versions? This was very recently implemented in #161 which doesn't prefix the path with `COLCON_CURRENT_PREFIX` if the following part is absolute.

PR #166, opened 2019-02-23
add new extension point to collect prefix paths

dirk-thomas:
This will enable that other prefix path ca be contributed which are not in the `COLCON_PREFIX_PATH` environment variable.


dirk-thomas:
Also register the extension point itself: ff5e2f7e0f5e98a972e3c59bdf33a8031e0aba3e.

PR #179, opened 2019-04-03
fix sourcing empty workspace built with `--merge-install`

mikaelarguedas:
Before this patch:
```
$ docker run -it --rm osrf/ros2:nightly
mkdir -p /tmp/foo/src && cd /tmp/foo
colcon build --merge-install
source install/setup.bash

  File "/tmp/foo/install/_local_setup_util.py", line 137, in <module>
    main()
  File "/tmp/foo/install/_local_setup_util.py", line 18, in main
    packages = get_packages(Path(__file__).parent, args.merged_install)
  File "/tmp/foo/install/_local_setup_util.py", line 42, in get_packages
    for p in (prefix_path / subdirectory).iterdir():
  File "/usr/lib/python3.6/pathlib.py", line 1081, in iterdir
    for name in self._accessor.listdir(self):
  File "/usr/lib/python3.6/pathlib.py", line 387, in wrapped
    return strfunc(str(pathobj), *args)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/foo/install/share/colcon-core/packages'
```

@karsten1987 FYI

### DIFF
--- a/colcon_core/shell/template/prefix_util.py
+++ b/colcon_core/shell/template/prefix_util.py
@@ -38,6 +38,9 @@ def get_packages(prefix_path, merged_install):
     # must match colcon_core.location.get_relative_package_index_path()
     subdirectory = 'share/colcon-core/packages'
     if merged_install:
+        # return if workspace is empty
+        if not (prefix_path / subdirectory).is_dir():
+            return packages
         # find all files in the subdirectory
         for p in (prefix_path / subdirectory).iterdir():
             if not p.is_file():
--- a/test/test_shell_template_prefix_util.py
+++ b/test/test_shell_template_prefix_util.py
@@ -29,6 +29,10 @@ def test_get_packages():
     with TemporaryDirectory(prefix='test_colcon_') as prefix_path:
         prefix_path = Path(prefix_path)
 
+        # mock no packages in not merged install layout
+        packages = get_packages(prefix_path, False)
+        assert packages == {}
+
         # mock packages in not merged install layout
         subdirectory = get_relative_package_index_path()
         for pkg_name in ('pkgA', 'pkgB'):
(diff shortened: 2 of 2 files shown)


colcon_core/shell/template/prefix_util.py
```diff
@@ -38,6 +38,9 @@ def get_packages(prefix_path, merged_install):
     # must match colcon_core.location.get_relative_package_index_path()
     subdirectory = 'share/colcon-core/packages'
     if merged_install:
+        # return if workspace is empty
+        if not (prefix_path / subdirectory).is_file():
```
dirk-thomas: Isn't this condition always going to be `True` since the tested path is supposed to be a directory? Shouldn't it be checking `is_dir()` instead?

mikaelarguedas: yes, already fixed

dirk-thomas:
Can you please update the tests to cover this particular case.

mikaelarguedas:
> Can you please update the tests to cover this particular case.

:+1: added test coverage for both merged and not merged layout in https://github.com/colcon/colcon-core/pull/179/commits/9357b5d68105387b2fad65983af5fce9dc09ad76.

test/test_shell_template_prefix_util.py
```diff
@@ -29,6 +29,10 @@ def test_get_packages():
     with TemporaryDirectory(prefix='test_colcon_') as prefix_path:
         prefix_path = Path(prefix_path)
 
+        # mock no packages in not merged install layout
```
dirk-thomas: Please remove "mock" from the comment since there is no mock involved in the following block.

Same below.

mikaelarguedas: done in https://github.com/colcon/colcon-core/pull/179/commits/dce2207141bfee1aceecce9806effed2f0e98ee7

dirk-thomas:
Thanks for the patch.

PR #180, opened 2019-04-08
avoid duplicate prefix paths

dirk-thomas:
Addresses part of colcon/colcon-ros#58.


mikaelarguedas:
Works for my use case :+1:

Didnt look into cases where skipping duplicates could impact the order of the resulting paths

PR #188, opened 2019-05-17
add colcon __main__ module so user can 'python -m colcon'

### LINKED ISSUE #181: Colcon runs in global interpreter
The colcon executable installed through apt has a shebang of `#!/usr/bin/python3`.
This prevents use in different virtual environments, especially since this shebang then puts the global python shebang into scripts built with colcon.

One (or more) of the following would make this easier:
- Recommended to install colcon-core locally (instead of globally with apt as per https://colcon.readthedocs.io/en/released/user/installation.html).
- Change the shebang on colcon to `/usr/bin/env python3`.
- Colcon should respect the Python interpreter specified with `$COLCON_PYTHON_EXECUTABLE` when building other python projects.

```
➜  ros2_ws git:(master) ✗ cat `which colcon`
#!/usr/bin/python3
# EASY-INSTALL-ENTRY-SCRIPT: 'colcon-core==0.3.20','console_scripts','colcon'
__requires__ = 'colcon-core==0.3.20'
import re
...
```

dirk-thomas:
Replaces #185. Closes #181.

### DIFF
--- a/colcon/__main__.py
+++ b/colcon/__main__.py
@@ -1,0 +1,11 @@
+# Copyright 2010 Dirk Thomas
+# Licensed under the Apache License, Version 2.0
+
+# This module exists for multi-Python environments
+# so the user can invoke colcon as `python -m colcon`
+
+import sys
+
+import colcon_core.command
+
+sys.exit(colcon_core.command.main())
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Coverage back to +4 of the +4 new lines: https://codecov.io/gh/colcon/colcon-core/pull/188/commits (not sure why the report above didn't get updated).

PR #201, opened 2019-07-16
use monotonic clock for duration

dirk-thomas:
Making the duration values robust to clock jumps.


clalancette:
I've run into this before, so this is welcome.  Looks good to me.

dirk-thomas:
> I've run into this before, so this is welcome.

If that is the case please report the problem right away (or even better create a pull request for it). Thanks.

PR #212, opened 2019-08-23
Update release suites for deb releases.

nuclearsandwich:


dirk-thomas:
Neither of the `colcon` packages has ever been released into non-LTS Ubuntu versions. So I am not sure these three should be added - especially since no ROS version targets any of these distributions.

nuclearsandwich:
I think it's reasonable to say that colcon is as necessary for building ROS 2 from source on a platform as rosdep or rosinstall_generator. It makes sense to me that if we're providing those tools on these platforms we ought to provide colcon as well.

dirk-thomas:
Since `cosmic` is already EOL I would rather not add it.

Regarding `disco` and `eoan` in my personal opinion it doesn't make sense to add them. The support window for each is just so short (9 months) that I don't think the overhead of adding new non-LTS distros to every package and removing them again when they become EOL is justified for the very limited user base of these distros. (This means every three months the list of suites needs to be touched assuming we don't want to keep EOL platforms in the list).

In the past we have not even created apt repos for these new non-LTS distros by the time they were already EOLed. Therefore I would argue those should rather use the available PIP packages instead.

dirk-thomas:
> I don't think the overhead of adding new non-LTS distros to every package and removing them again when they become EOL is justified for the very limited user base of these distros. (This means every three months the list of suites needs to be touched assuming we don't want to keep EOL platforms in the list).

If others disagree on this and want these platforms be added I would like to get a commitment that someone will also take on the extra effort every three months to update the suites accordingly.

nuclearsandwich:
> I don't think the overhead of adding new non-LTS distros to every package and removing them again when they become EOL is justified for the very limited user base of these distros.

I agree with you. But when I raised that as an argument for not creating ROS repos for the non-LTS releases at all it was dismissed since there may be users who want the tools on those releases. So my question is what makes colcon different from any of the other packages required to produce ROS?

dirk-thomas:
> So my question is what makes colcon different from any of the other packages required to produce ROS?

`colcon` is living outside of ROS and I am kind of the only person from the ROS team working on / maintaining it. Therefore I am looking to keep the overhead as low as possible.

> But when I raised that as an argument for not creating ROS repos for the non-LTS releases at all it was dismissed since there may be users who want the tools on those releases.

We could revisit the policy for ROS packages. But that would need to be posted to the community for feedback. Releasing into every distro when it becomes available was simply what we did until recently (when every Ubuntu distro also had a ROS distro targeting it). With the larger gap between Melodic and Noetic it is the first time the case that Ubuntu 18.10 - 19.10 aren't targeted by any ROS 1 distro.

mikaelarguedas:
Came from #253 and was redirected here.

As a user it's more convenient to have the apt packages to benefit from updates more easily. While being less convenient using pip is totally fine as well.

My only preference would be to get out of the current state as it is confusing that packages are available but too outdated to be able to build ROS 2. So either pip and purging the repos from the python3-colcon-* or releasing them in non-LTS distros.

---

From a technical point of view it seems that if adding the new distros is done for all ROS packages anyway, it should be minimal overhead to have a script that updates ROS and colcon files and releases them

nuclearsandwich:
> `colcon` is living outside of ROS and I am kind of the only person from the ROS team working on / maintaining it. Therefore I am looking to keep the overhead as low as possible.

This is reasonable. As much as I would like to join you in colcon maintenance in order to deliver it on other platforms, I don't have my head above water to meet that commitment. I'll close these PRs and remove the included colcon packages from the repositories.

nuclearsandwich:
The colcon packages imported during the creation of the cosmic, disco, and eoan bootstrap repositories have been removed and that has been propagated out to all the ROS repositories.

PR #222, opened 2019-09-06
[Windows] Use default expansion for _colcon_python_executable

seanyen:
It seems to be a regression after https://github.com/colcon/colcon-core/pull/209, where it removed `enabledelayedexpansion` for `_colcon_run_ordered_commands` but forgot to revert the delay expansion usage. As a result, whenever the generated `local_setup.bat` runs, it throws out `error: unable to find python executable` even the Python executable exists.


seanyen:
cc @dirk-thomas

dirk-thomas:
Thank you for the fix.

PR #232, opened 2019-09-18
add COLCON_WARNINGS environment variable

dirk-thomas:
This environment variable provide a way to specify warnings filter which are scoped to modules matching the regular expression `colcon.*`. Using `PYTHONWARNINGS` directly isn't possible since it doesn't support using a regular expression: see https://bugs.python.org/issue34624.

### DIFF
--- a/colcon_core/command.py
+++ b/colcon_core/command.py
@@ -10,20 +10,57 @@ import signal
 import sys
 import traceback
 
-from colcon_core.argument_parser import decorate_argument_parser
-from colcon_core.argument_parser import SuppressUsageOutput
-from colcon_core.entry_point import load_entry_points
 from colcon_core.environment_variable import EnvironmentVariable
-from colcon_core.location import create_log_path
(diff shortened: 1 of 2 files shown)


colcon_core/command.py
```diff
@@ -10,20 +10,57 @@
...
+            print(
+                'The module field of the {WARNINGS_ENVIRONMENT_VARIABLE.name} '
+                'filter should be empty, otherwise use PYTHONWARNINGS instead'
+                .format_map(locals()), file=sys.stderr)
+            sys.exit(1)
+        warnings.filterwarnings(
+            action, message=message, category=category or Warning,
+            module='colcon.*', lineno=line if line else 0)
+
+from colcon_core.argument_parser import decorate_argument_parser  # noqa: E402 E501 I100 I202
```
jacobperron: Just to clarify, these imports must go after in order to detect warnings those modules?

dirk-thomas: Correct.

PR #234, opened 2019-09-18
close file handle after use

dirk-thomas:
Detected using this new feature: #232.

### DIFF
--- a/colcon_core/subprocess.py
+++ b/colcon_core/subprocess.py
@@ -231,6 +231,7 @@ def _blocking_fd2callback(stream, callback):
             # this is how the fd signals the EOF
             break
         callback(line.encode())
+    stream.close()
 
 
 async def _pipe2callback(stream, callback, other_stream=None):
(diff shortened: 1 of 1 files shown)


rotu:
Generally, resources should be disposed of in the scope that created them and disposed even if an exception was encountered.

I would move `os.fdopen` into `_fd2callback` and use the context manager protocol to ensure it gets closed:

```
async def _fd2callback(fd, callback):
  """Coroutine reading from fd and invoking the callback for each line."""
  with open(fd) as stream:
    func = partial(_blocking_fd2callback, stream, callback)
    loop = asyncio.get_event_loop()
    await loop.run_in_executor(None, func)
```

dirk-thomas:
@rotu Very good point. Please see 29a419d33600c1723db06e1635192bea6c5645a4 which moves the creation of the file handle into the `_blocking_fd2callback` function and uses a context manager there.

PR #235, opened 2019-09-26
rename rc -> completed

rotu:
I assume rc stood for "return code", but it's a CompletedProcess object.
Also, don't check if check_call returns None. It never does.

### DIFF
--- a/colcon_core/task/python/build.py
+++ b/colcon_core/task/python/build.py
@@ -56,9 +56,9 @@ class PythonBuildTask(TaskExtensionPoint):
             env.get('PYTHONPATH', '')
 
         if not args.symlink_install:
-            rc = await self._undo_develop(pkg, args, env)
-            if rc and rc.returncode:
-                return rc.returncode
+            none_or_completed = await self._undo_develop(pkg, args, env)
+            if none_or_completed and none_or_completed.returncode:
+                return none_or_completed.returncode
 
             # invoke `setup.py install` step with lots of arguments
             # to avoid placing any files in the source space
@@ -73,9 +73,10 @@ class PythonBuildTask(TaskExtensionPoint):
                 '--single-version-externally-managed',
             ]
             self._append_install_layout(args, cmd)
-            rc = await check_call(self.context, cmd, cwd=args.path, env=env)
-            if rc and rc.returncode:
-                return rc.returncode
+            completed = await check_call(
+                self.context, cmd, cwd=args.path, env=env)
+            if completed.returncode:
+                return completed.returncode
 
         else:
             self._undo_install(pkg, args, setup_py_data, python_lib)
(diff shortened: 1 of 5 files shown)


colcon_core/task/python/build.py
```diff
@@ -125,10 +126,9 @@ def __init__(self):  # noqa: D107
                 '--uninstall', '--editable',
                 '--build-directory', os.path.join(args.build_base, 'build')
             ]
-            rc = await check_call(
+            completed = await check_call(
                 self.context, cmd, cwd=args.build_base, env=env)
-            if rc:
-                return rc
+            return completed
```
dirk-thomas: In this last case the variable `completed` isn't necessary anymore. Instead it can directly `return await check_call(...)`.

rotu:
> It seems there are many more locations with the same pattern.

I'm just making the changes in this one file because I found myself correcting these instances in another PR against this file and knew you'd complain if I made them there. Would you like a bigger PR? Because something called `check_call` should really be throwing an exception on a failed process, to be consistent with https://docs.python.org/3.7/library/subprocess.html#subprocess.check_call.

dirk-thomas:
> I'm just making the changes in this one file because I found myself correcting these instances in another PR against this file and knew you'd complain if I made them there.

:+1:

> Would you like a bigger PR?

How about we keep this open until a few other pending PRs have landed and then we update all instances at once? I can do that follow up too if you like...

> Because something called `check_call` should really be throwing an exception on a failed process

That would break API behavior which I don't think is necessary.

rotu:
> That would break API behavior which I don't think is necessary.

Is this part of the public API? If so, it should *definitely* be changed as it subverts expectations. Where else is it used?

dirk-thomas:
> Is this part of the public API?

Any API not explicitly "declared" as private using a leading underscore is public in Python.

> If so, it should *definitely* be changed as it subverts expectations to not check the result of a call in a function called `check_call`.

I understand your point of view but I any kind of potential breakage due to this "clean up" isn't acceptable imo.

Maybe we introduce a new API beside the existing one and modify existing code to use it instead?

> Where else is it used?

Please search the code of the various `colcon` repositories. Beside `colcon-core` it is used e.g. in `colcon-cmake` and I expect in any kind of extension providing support for a different build system.

rotu:
> I understand your point of view but I any kind of potential breakage due to this "clean up" isn't acceptable imo.

Agreed.

> Maybe we introduce a new API beside the existing one and modify existing code to use it instead?

:+1: that's probably the cleanest approach

rotu:
rebased on master

dirk-thomas:
I rebased the patch as well as updated it a bit (some additional changes, some changes to return values for internal methods). This should now cover all the cases where `CompletedProcess` is returned.

PR #236, opened 2019-09-26
redirect pytest warnings summary to stderr

dirk-thomas:
With the goal to make them more visible. A package can always opt to filter the warnings using e.g. the `filterwarnings` option in a `pytest.ini` file.

### DIFF
--- a/colcon_core/task/python/test/pytest.py
+++ b/colcon_core/task/python/test/pytest.py
@@ -1,11 +1,15 @@
-# Copyright 2016-2018 Dirk Thomas
+# Copyright 2016-2019 Dirk Thomas
 # Licensed under the Apache License, Version 2.0
 
 import os
 from pathlib import Path
 from pathlib import PurePosixPath
 import sys
+import traceback
+import types
 
+from colcon_core.event.output import StderrLine
+from colcon_core.event.output import StdoutLine
 from colcon_core.event.test import TestFailure
 from colcon_core.plugin_system import satisfies_version
 from colcon_core.plugin_system import SkipExtensionException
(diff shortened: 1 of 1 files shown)


rotu:
Why reparse stdout instead of making sure these messages get sent to the correct stream in the first place?

dirk-thomas:
> Why reparse stdout instead of making sure these messages get sent to the correct stream in the first place?

Do you have a suggestion how to achieve that?

rotu:
either with the `pytest_runtest_logreport`, `pytest_terminal_summary` or `pytest_warning_captured` hook. 

https://docs.pytest.org/en/latest/_modules/_pytest/hookspec.html

dirk-thomas:
Closing in favor of #338.

tylerjw:
Sadly it looks like the ninja project itself considers "feature" to be not-useful: https://github.com/ninja-build/ninja/issues/1537

Their reasoning is that the error message is not useful without the context of what command caused it and therefore if you care about why it failed, you also care about what command was called when it failed.

tylerjw:
The suggested solution there is this:

```bash
NINJA_STATUS="[ninja][%f/%t]
filter() {
  local l n
  while read -r l; do
    n=${l#\[ninja\]}
    [ "x$l" != "x$n" ] && printf "%s\n" "$n" || printf "%s\n" "$l" >&2
  done
}

NINJA_STATUS="[ninja][%f/%t] " ninja | filter
```

I don't know how difficult this would be to implement (here) but it is simple enough.

dirk-thomas:
@tylerjw Your comments seem to be off topic for this ticket. colcon/colcon-cmake#67 seems to be discussing what you are referring to. Please consider to contribute a pull request to `colcon-cmake` to implement custom logic to separate the Ninja output.

tylerjw:
I'm sorry about that... I'll re-target my comments.

PR #247, opened 2019-10-16
Add support for prepending multiple values to env variables from .dsv files

jacobperron:
To add context, when working with Java projects it can happen that a package produces multiple jar files that should be added to the `CLASSPATH` environment variable.
Example .dsv file for `rcljava_common`:

```
prepend-non-duplicate;CLASSPATH;share/rcljava_common/java/commons-lang3-3.7.jar;share/rcljava_common/java/slf4j-api-1.7.21.jar;share/rcljava_common/java/log4j-1.2.17.jar;share/rcljava_common/java/slf4j-log4j12-1.7.21.jar;share/rcljava_common/java/slf4j-jdk14-1.7.21.jar;share/rcljava_common/java/rcljava_common.jar
```

### DIFF
--- a/colcon_core/shell/template/prefix_util.py.em
+++ b/colcon_core/shell/template/prefix_util.py.em
@@ -256,22 +256,25 @@ def handle_dsv_types_except_source(type_, remainder, prefix):
         DSV_TYPE_PREPEND_NON_DUPLICATE,
         DSV_TYPE_PREPEND_NON_DUPLICATE_IF_EXISTS
     ):
-        env_name, value = remainder.split(';', 1)
-        if not value:
-            value = prefix
-        elif not os.path.isabs(value):
-            value = os.path.join(prefix, value)
-        if (
-            type_ == DSV_TYPE_PREPEND_NON_DUPLICATE_IF_EXISTS and
-            not os.path.exists(value)
-        ):
-            comment = 'skip extending {env_name} with not existing path: ' \
-                '{value}'.format_map(locals())
-            if _include_comments():
-                commands.append(
-                    FORMAT_STR_COMMENT_LINE.format_map({'comment': comment}))
-        else:
-            commands += _prepend_unique_value(env_name, value)
+        env_name_and_values = remainder.split(';')
+        env_name = env_name_and_values[0]
+        values = env_name_and_values[1:]
+        if 0 == len(values):
+            values.append(prefix)
+        for value in values:
+            if not os.path.isabs(value):
+                value = os.path.join(prefix, value)
+            if (
+                type_ == DSV_TYPE_PREPEND_NON_DUPLICATE_IF_EXISTS and
+                not os.path.exists(value)
+            ):
+                comment = 'skip extending {env_name} with not existing path: ' \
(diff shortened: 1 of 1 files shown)


colcon_core/shell/template/prefix_util.py.em
```diff
@@ -256,22 +256,25 @@ def handle_dsv_types_except_source(type_, remainder, prefix):
...
-            if _include_comments():
-                commands.append(
-                    FORMAT_STR_COMMENT_LINE.format_map({'comment': comment}))
-        else:
-            commands += _prepend_unique_value(env_name, value)
+        env_name_and_values = remainder.split(';')
+        env_name = env_name_and_values[0]
+        values = env_name_and_values[1:]
+        if 0 == len(values):
+            values.append(prefix)
```
dirk-thomas: Since the `.dsv` file always contains a trailing semicolon if there are no actaul path values the `values` variable is never empty but always at least contains one empty string. So this logic doesn't make sense.

colcon_core/shell/template/prefix_util.py.em
```diff
@@ -256,22 +256,25 @@ def handle_dsv_types_except_source(type_, remainder, prefix):
...
+        for value in values:
+            if not os.path.isabs(value):
```
dirk-thomas: This condition needs to check for an empty `value` as it did before.

colcon_core/shell/template/prefix_util.py.em
jacobperron: Guarded against empty strings in the list d6fdbd5

colcon_core/shell/template/prefix_util.py.em
jacobperron: Since there are no empty strings in the list (d6fdbd5) I don't think we need to check for empty values.

colcon_core/shell/template/prefix_util.py.em
dirk-thomas: This is still wrong. The list can contain an empty value or multiple values where one could be empty. They must not be removed since they represent the prefix path.

jacobperron: Oh, I thought that we'd only append the prefix path in the case there are no values provided.
I'll change the logic so every empty value is set to the prefix path.

Where is is documented that an empty value represents the prefix path (or that the prefix path is appended)? I think it might need to be updated along with this change.

dirk-thomas: All values are simply either relative paths or absolute path. And relative paths are prefixed with the current prefix path to make them absolute.

jacobperron: Might be nice to document the .dsv feature altogether somewhere, so it's easier for people to leverage it if as they need. But this is outside the scope of this PR.

I've updated again, using the prefix path in place of all empty values (c8b8dac).

colcon_core/shell/template/prefix_util.py.em
jacobperron: See c8b8dac

PR #249, opened 2019-10-18
support more characters in environment variable names on Windows

### LINKED ISSUE #248: Colcon shell.get_environment_variables does not correctly handle some Windows environment variables
When colcon_core.shell.get_environment_variables() is decoding the environment variable string, it makes invalid assumptions about the format of a valid environment variable key. Specifically, it uses the regular expression: '^[a-zA-Z0-9_%]+$' (around line 328). This does not correctly handle the environment variable strings `PROGRAMFILES(X86)` (cmd) or `ProgramFiles(x86)` (PowerShell). Instead, this environment variable and its key are appented to the previous environment variable, pontetially invalidating that variable.

This can result in a polluted environment causing downstream issues. See https://github.com/colcon/colcon-cmake/issues/54

The expected behaviour is that `PROGRAMFILES(X86)` is stored as its own key/value pair in the resulting environment dictionary and other valid variables are not polluted.

One potential fix is to use a different regular expression for Windows, allowing '(' and ')' characters.

dirk-thomas:
Fixes #248 

On Windows any of the following characters is allowed in an environment variable name (to my knowledge):
* lower alpha
* upper alpha
* digit
* underscore
* and the special characters not allowed on non-Windows:
  * percent (for dynamic names I guess)
  * `(){}[]$*+-\/"#',;.@!?`

For non-Windows the patch removes the `%` character (which was only added for Windows in the first place) as well as enforce that the first character is not a digit.

### DIFF
--- a/colcon_core/shell/__init__.py
+++ b/colcon_core/shell/__init__.py
@@ -7,6 +7,7 @@ import locale
 import os
 from pathlib import Path
 import re
+import sys
 import traceback
 import warnings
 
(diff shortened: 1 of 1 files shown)


colcon_core/shell/__init__.py
```diff
@@ -325,7 +326,13 @@ def get_shell_extensions():
                 "encoding '{encoding}': {line_replaced}".format_map(locals()))
             continue
         parts = line.split('=', 1)
-        if len(parts) == 2 and re.match('^[a-zA-Z0-9_%]+$', parts[0]):
+        if sys.platform != 'win32':
+            regex = '^[a-zA-Z_][a-zA-Z0-9_]*$'
+        else:
+            regex = '^[a-zA-Z0-9%' + ''.join(
+                '\\' + c for c in r'_(){}[]$*+-\/"#\',;.@!?'
```
jacobperron: Does the underscore need to be escaped or can it be moved to the first part of the expression?

dirk-thomas: I doesn't need to be escaped: 4333f6e00cbe35f716c27fd98e4aa4d985e83140.

PR #251, opened 2019-10-21
check that no arguments are passed to the setup() function in setup.py

dirk-thomas:
Prevents reading only the `setup.cfg` file when the `setup()` call in the `setup.py` file contains more information.

In this case the [colcon-python-setup-py](https://github.com/colcon/colcon-python-setup-py) should be used to identify the Python and collect its information.

### DIFF
--- a/colcon_core/package_identification/python.py
+++ b/colcon_core/package_identification/python.py
@@ -90,10 +90,10 @@ def is_reading_cfg_sufficient(setup_py):
     Check the content of the setup.py file.
 
     If the ``setup()`` function is called with no arguments or only a
-    ``cmdclass`` is is sufficient to only read the content of the ``setup.cfg``
+    ``cmdclass`` it is sufficient to only read the content of the ``setup.cfg``
     file.
 
-    :param setup_cfg: The path of the setup.py file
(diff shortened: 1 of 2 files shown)


colcon_core/package_identification/python.py
```diff
@@ -72,6 +85,26 @@ def getter(env):
         desc.metadata['get_python_setup_options'] = getter
 
 
+def is_reading_cfg_sufficient(setup_py):
+    """
+    Check the content of the setup.py file.
+
+    If the ``setup()`` function is called with no arguments or only a
+    ``cmdclass`` is is sufficient to only read the content of the ``setup.cfg``
```
jacobperron:
```suggestion
    ``cmdclass`` it is sufficient to only read the content of the ``setup.cfg``
```

colcon_core/package_identification/python.py
```diff
@@ -72,6 +85,26 @@ def getter(env):
...
+    file.
+
+    :param setup_cfg: The path of the setup.py file
```
jacobperron:
```suggestion
    :param setup_py: The path of the setup.py file
```

PR #259, opened 2019-10-25
avoid repeated deepcopy of descriptors during identification

dirk-thomas:
Refactor the always copied descriptors which are supposed to be updated by an extension to reuse the copied descriptor in the next cycle if it hasn't been changed. That avoids a `deepcopy` of the descriptor in the majority of the cases.

For an example ROS 2 workspace with 268 pkgs the time to list goes down by roughly 10% (from 1.7s to 1.55s).

### DIFF
--- a/colcon_core/package_identification/__init__.py
+++ b/colcon_core/package_identification/__init__.py
@@ -119,17 +119,28 @@ def identify(
     return None
 
 
+# the following variable only exists to avoid repeatedly copying descriptors
+_reused_descriptor_instance = None
+
+
 def _identify(extensions_same_prio, desc):
+    global _reused_descriptor_instance
     logger.log(
         1, '_identify(%s) by extensions %s',
         desc.path, sorted(extensions_same_prio.keys()))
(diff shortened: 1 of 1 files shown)


colcon_core/package_identification/__init__.py
```diff
@@ -164,3 +180,10 @@ def _identify(extensions_same_prio, desc):
         "Package '{desc.path}' with type '{desc.type}' and name '{desc.name}'"
         .format_map(locals()))
     return desc
+
+
+def _are_descriptors_equal(desc1, desc2):
```
jacobperron: Why not use the existing `__eq__` definition? I notice it only uses `path`, `type`, and `name` slots, maybe it should be updated to include the remaining slots?

dirk-thomas: The `__eq__` only checks if the object represents the same package (which is uniquely identifies by these three attributes). If additional dependencies and metadata are added they are not used for that comparison. Therefore I needed a separate comparison covering all slots.

PR #260, opened 2019-10-25
add minimum version for flake8 to provide extend_ignore

rotu:
Require 3.6.0 for extend_ignore


dirk-thomas:
Thanks for the patch.

Ref to the change log: https://flake8.pycqa.org/en/latest/release-notes/3.6.0.html#features

I also updated the title to be more descriptive since it will become the commit message after being squash merged.

PR #261, opened 2019-10-25
fix command to remove trailing separator

dirk-thomas:
Follow up of #254 fixing the logic in sh/bash.


jacobperron:
I guess `test` is not meant to be used with `if`. The equivalent would use `&&`:

    test "$(echo -n ${name} | tail -c 1)" = ":" && export {name}=${{{name}%?}}

jacobperron:
Now I see it... `then;` should have a trailing semi-colon :upside_down_face:

PR #276, opened 2019-12-03
fixed KeyError in _prepend_unique_value

tobiasblass:
_prepend_unique_value fails to initialize env_state if the environment
variable is the empty string. Instead, it should treat missing and
empty variables identically, just like the shell does.

### DIFF
--- a/colcon_core/shell/template/prefix_util.py.em
+++ b/colcon_core/shell/template/prefix_util.py.em
@@ -295,10 +295,10 @@ env_state = {}
 def _prepend_unique_value(name, value):
     global env_state
     if name not in env_state:
-        if name not in os.environ:
-            env_state[name] = set()
         if os.environ.get(name):
             env_state[name] = set(os.environ[name].split(os.pathsep))
+        else:
+            env_state[name] = set()
     # prepend even if the variable has not been set yet, in case a shell script sets the
     # same variable without the knowledge of this Python script.
     # later _remove_trailing_separators() will cleanup any unintentional trailing separator


dirk-thomas:
Please provide reproducible steps which fail without / pass with this patch.

tobiasblass:
> Please provide reproducible steps which fail without / pass with this patch.

Sure, sorry about that. Just source the install script with one of the path variables empty, e.g.,
```
CMAKE_PREFIX_PATH='' . workspace/install/local_setup.bash
```

Without this patch, you get a KeyError, because `env_state['CMAKE_PREFIX_PATH']` is undefined. With this patch, `env_state['CMAKE_PREFIX_PATH']` is initialized to the empty set, just as if CMAKE_PREFIX_PATH was undefined.

colcon_core/shell/template/prefix_util.py.em
```diff
@@ -295,10 +295,10 @@ env_state = {}
 def _prepend_unique_value(name, value):
     global env_state
     if name not in env_state:
-        if name not in os.environ:
-            env_state[name] = set()
-        if os.environ.get(name):
+        if os.environ.get(name, ''):
```
dirk-thomas: The default `default` value of `get()` is `None` and is sufficient for this check. So no need to change this line.

tobiasblass: That's true, but I think this formulation makes it clearer that missing environment variables are treated as if they contained the empty string. However, this is a minor stylistic thing, so if you find it irritating I can remove the default value.

dirk-thomas: If I read the `, ''` I am asking myself why is it there? Why is the default value `None` not sufficient. Since there is no need for it I rather don't want to think about it :wink:

Thanks for updating it.

PR #279, opened 2019-12-09
add missing quote around set arguments in bat

dirk-thomas:


hidmic:
Not just yet, getting:

```bat
if "%PYTHONPATH:~-1%"==";" set "PYTHONPATH=%PYTHONPATH:~0,-1%"
'if' is not recognized as an internal or external command,
operable program or batch file.
```

with these changes.

dirk-thomas:
@hidmic Please try again with the updated patch.

jacobperron:
Approval under the assumption this fixes things (I haven't given it a try yet).

dirk-thomas:
> I haven't given it a try yet.

I can wait for that - for this part it's better to double/triple check.

hidmic:
Took me a while to get back to this, but I can confirm the patch works just fine. Thanks!

PR #280, opened 2019-12-09
update constructor docblocks to pass pydocstyle 5.x checks

dirk-thomas:


colcon_core/event/command.py
```diff
@@ -93,7 +93,7 @@ class CommandEnded(Command):
 
     def __init__(self, cmd, *, cwd, returncode, env=None, shell=False):
         """
-        Constructor.
+        Construct a CommandEnded.
```
nuclearsandwich: Are you concerned with whether the docstring is grammatical? For example `Construct a CommandEnded.` doesn't read easily in English since CommandEnded isn't clearly a noun. When updating the docstrings for launch I had used names from the class hierarchy to build something which sounds like a valid subject in English. (example: https://github.com/ros2/launch/pull/362/files#diff-9c41e31fadb67a719d012697247b072aR28).

dirk-thomas: Since the whole `CommandEnded` is a single name I would think this is fine: like "Construct a Foo". I don't mind enough to come up with case specific formulations since it doesn't add any documentation value.

PR #281, opened 2019-12-09
Add missing quotation mark

jacobperron:
Otherwise, calling setup.bat will result in multiple errors like the following:

    " set CMAKE_PREFIX_PATH=C:\J\workspace\packaging_windows\ws\install
    was unexpected at this time.

I guess some users may not have noticed if they were still using an
older version of colcon-core.

### DIFF
--- a/colcon_core/shell/bat.py
+++ b/colcon_core/shell/bat.py
@@ -26,7 +26,7 @@ class BatShell(ShellExtensionPoint):
     FORMAT_STR_USE_ENV_VAR = '%{name}%'
     FORMAT_STR_INVOKE_SCRIPT = 'call:_colcon_prefix_bat_call_script ' \
         '"{script_path}"'
-    FORMAT_STR_REMOVE_TRAILING_SEPARATOR = 'if "%{name}:~-1%==";" ' \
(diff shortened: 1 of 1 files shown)


colcon_core/shell/bat.py
```diff
@@ -26,7 +26,7 @@ class BatShell(ShellExtensionPoint):
     FORMAT_STR_USE_ENV_VAR = '%{name}%'
     FORMAT_STR_INVOKE_SCRIPT = 'call:_colcon_prefix_bat_call_script ' \
         '"{script_path}"'
-    FORMAT_STR_REMOVE_TRAILING_SEPARATOR = 'if "%{name}:~-1%==";" ' \
+    FORMAT_STR_REMOVE_TRAILING_SEPARATOR = 'if "%{name}:~-1%"==";" ' \
         'set {name}=%{name}:~0,-1%'
```
dirk-thomas: I would think the `name=value` pair should still be quotes as in #279?

jacobperron:
To summarize, we realized that this solution won't work since the format string ultimately gets run with `call` (https://github.com/colcon/colcon-core/blob/8a06f55906938ac97da8072ce9ae52fb9fb9e9cf/colcon_core/shell/template/prefix.bat.em#L87). See #279 for an alternative solution.

PR #286, opened 2020-01-06
add more Python versions to the test matrix

dirk-thomas:
* On Linux check: Python 3.5 - 3.8 (all used by actively supported distros)
* On Windows (since it is a rolling release) only check the latest: Python 3.8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,12 @@
 language: python
-python: "3.5"
-dist: trusty
-sudo: true
+dist: xenial
+sudo: false
+python:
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Needs a bug fix release of [pydocstyle](https://github.com/PyCQA/pydocstyle/pull/445) to pass CI.

dirk-thomas:
Also added Python 3.9 for Linux and use it (since it is the latest release) on Windows.

jacobperron (approved):

Issue #274, opened 2019-12-03
Confusing interaction between --base-paths and --paths


rotu:
When searching for one package by path using `colcon info --paths /path/to/package`, colcon performed recursive package discovery. This seems to be because I set my `base-paths` in the defaults yaml file. Passing `--base-path "."` makes colcon only discover the package as specified.

defaults.yaml:
```
...
info:
  base-paths: [src]
```

`colcon list --paths-only`

> /opt/ros/master/src/ApexAI/performance_test/performance_test
> /opt/ros/master/src/ament/ament_cmake/ament_cmake
> ...

`colcon list --paths-only --paths src/ros2/rmw/rmw`

> /opt/ros/master/src/ApexAI/performance_test/performance_test
> /opt/ros/master/src/ament/ament_cmake/ament_cmake
> ...

` colcon list --paths-only --paths src/ros2/rmw/rmw --base-paths .`
> /opt/ros/master
> /opt/ros/master/src/ros2/rmw/rmw
(even though I have a top-level CMakeLists.txt in the current directory, this is unexpected)

`colcon list --paths-only --paths src/ros2/rmw/rmw --base-paths ""`
`colcon list --paths-only --paths src/ros2/rmw/rmw --base-paths ''`
`colcon list --paths-only --paths src/ros2/rmw/rmw --base-paths`
> /opt/ros/master/src/ros2/rmw/rmw

This is the expected result.

It seems that passing `--paths` on the command line should cause colcon to not recurse through base-paths in the yaml file. Better yet, have no special handling here and set the default '--paths' value to `"**"`

dirk-thomas:
Conceptionally the discovery extensions distinguish between the ones which have explicit parameters from extensions which only have a default value. At the moment the `default=...` value of `argparse` arguments can't be distinguished from arguments which have been explicitly passed through the CLI. That is why the recursive crawling (`--base-paths`) happens even though explicit paths are passed with `--paths`.

> It seems that passing `--paths` on the command line should cause colcon to not recurse through base-paths in the yaml file.

The solution to that problem is to be able to identify a default value independently from the actual value. Please see #288 and colcon/colcon-recursive-crawl#17 which add this for the discovery extensions. Additionally the `colcon-default` package needs to annotate the default values provided from a config file in the same way (see colcon/colcon-default#20).

Those changes together should provide the expected behavior. It would be great if you could give those patches a try and comment here with the result.

> Better yet, have no special handling here and set the default '--paths' value to "**"

[...]

PR #288, opened 2020-01-07
wrap default value to distinguish from explicitly passed arguments

dirk-thomas:
Together with the referenced PRs this addresses #274.

Add concept of wrapped default values in order to distinguish them from explicitly passed arguments. The wrapped values aim to have as similar behavior as possible to the original value type.

### DIFF
--- a/colcon_core/argument_default.py
+++ b/colcon_core/argument_default.py
@@ -49,7 +49,6 @@ def wrap_default_value(value):
     if is_default_value(value):
         raise ValueError(
             'the passed value is already wrapped: ' + str(type(value)))
-        return value
     if type(value) in _types:
         return _types[type(value)](value)
     return value
--- a/colcon_core/argument_default.py
+++ b/colcon_core/argument_default.py
@@ -1,0 +1,60 @@
+# Copyright 2020 Dirk Thomas
+# Licensed under the Apache License, Version 2.0
+
+
+class BooleanDefaultValue:
+    """Boolean value distinguishable from explicitly passed arguments."""
+
+    def __init__(self, value):  # noqa: D107
+        assert isinstance(value, bool)
+        self._value = value
+
+    def __bool__(self):  # noqa: D105
+        return self._value
+
+
+class ListDefaultValue(list):
+    """List value distinguishable from explicitly passed arguments."""
+
+    pass
+
+
+class StringDefaultValue(str):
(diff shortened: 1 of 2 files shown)


colcon_core/argument_default.py
```diff
@@ -0,0 +1,60 @@
...
+    The custom types try to match the behavior of the original type as much as
+    possible.
+
+    :param value: The default value to be wrapped
+    :returns: The wrapped value if the value type is supported, otherwise the
+      passed value
+    """
+    global _types
+    if is_default_value(value):
+        assert False
```
jacobperron: Why assert False?

colcon_core/package_discovery/path.py
```diff
@@ -30,13 +32,14 @@ def add_arguments(
             '--paths',
             nargs='*' if not single_path else '?',
             metavar='PATH',
-            default='.' if with_default else None,
+            default=wrap_default_value(['.']) if with_default else None,
```
jacobperron: Shouldn't we continue to pass a string, rather than a list?
```suggestion
            default=wrap_default_value('.') if with_default else None,
```

dirk-thomas: The argument uses `nargs='*'` which makes the value a list. Setting the default to a string was a bug before. Luckily iterating a string with a single character has the same effect as iterating a list of strings in this case. For a different value that would be broken (see e.g. colcon/colcon-metadata#18).

colcon_core/argument_default.py
dirk-thomas: A value which is already wrapped shouldn't be wrapped again. I switched the assertion into an exception for better feedback: fafb10d229236bcdac988c06abee16cf8ac82791.

colcon_core/argument_default.py
```diff
@@ -0,0 +1,61 @@
...
+
+    :param value: The default value to be wrapped
+    :returns: The wrapped value if the value type is supported, otherwise the
+      passed value
+    """
+    global _types
+    if is_default_value(value):
+        raise ValueError(
+            'the passed value is already wrapped: ' + str(type(value)))
+        return value
```
jacobperron: Since we're raising an exception on the previous line, this statement can be removed.

dirk-thomas: Good catch, fixed in 89688e9eb9830c6dd36c1c1f929a6adce1aa3a1d.

PR #292, opened 2020-01-13
add --pytest-with-coverage option

dirk-thomas:
Allows to force coverage information for Python packages using `pytest` even if they don't declare an explicit test dependency on `pytest-cov`.

### DIFF
--- a/colcon_core/task/python/test/pytest.py
+++ b/colcon_core/task/python/test/pytest.py
@@ -40,6 +40,10 @@ class PytestPythonTestingStep(PythonTestingStepExtensionPoint):
             help='Pass arguments to pytests. '
             'Arguments matching other options must be prefixed by a space,\n'
             'e.g. --pytest-args " --help"')
+        parser.add_argument(
+            '--pytest-with-coverage',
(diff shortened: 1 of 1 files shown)


colcon_core/task/python/test/pytest.py
```diff
@@ -40,6 +40,10 @@ def add_arguments(self, *, parser):  # noqa: D102
             help='Pass arguments to pytests. '
             'Arguments matching other options must be prefixed by a space,\n'
             'e.g. --pytest-args " --help"')
+        parser.add_argument(
+            '--pytest-with-coverage',
+            action='store_true',
+            help='Generate coverage information')
```
hidmic: @dirk-thomas nit^2: 
```suggestion
            help='Generate pytest coverage information')
```

dirk-thomas: The generated coverage data isn't specific to `pytest`. Therefore I didn't include it in the help text.

hidmic: That's a fair point, disregard.

PR #296, opened 2020-01-29
invoke pytest directly instead of through the setup.py file

dirk-thomas:
As of `setuptools` 41.5.0 invoking tests through the `setup.py` file is deprecated (see pypa/setuptools#1878):

> WARNING: Testing via this command is deprecated and will be removed in a future version. Users looking for a generic test entry point independent of test runner are encouraged to use tox.

This PR only changes the `pytest` based testing extension.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ env:
 install:
   - pip install -U setuptools
   # install_requires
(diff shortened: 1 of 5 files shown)


dirk-thomas:
Rebases after merging #297 and #298.

PR #305, opened 2020-02-20
add exclude argument to load_entry_points

dirk-thomas:
This allows avoiding to load undesired entry points.

### DIFF
--- a/bin/colcon
+++ b/bin/colcon
@@ -23,12 +23,14 @@ from colcon_core import entry_point  # noqa: E402
 custom_entry_points = {}
 
 
-def custom_load_entry_points(group_name):  # noqa: D103
+def custom_load_entry_points(group_name, *, exclude_names=None):  # noqa: D103
     global custom_entry_points
     assert group_name in custom_entry_points, \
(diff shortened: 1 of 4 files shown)


sloretz:
CI seems unhappy here

https://github.com/colcon/colcon-core/blob/b641351114484fa8a02afd9c6a36818f0a06749b/bin/colcon#L26-L35

bin/colcon
```diff
@@ -23,12 +23,14 @@ from colcon_core import entry_point  # noqa: E402
 custom_entry_points = {}
 
 
-def custom_load_entry_points(group_name):  # noqa: D103
+def custom_load_entry_points(group_name, *, exclude_names=None):  # noqa: D103
```
sloretz: The change looks fine, but I'm curious. Why does `bin/colcon` monkey patch `entry_point.load_entry_points`? What would happen if it used the original?

dirk-thomas: In thi bootstrap script `colcon` has not been installed yet so there are no entry points registered. This logic provide the entry points to bootstrap `colcon`. Once `colcon` is installed this script isn't used anymore (see the comment at the top of the file).

PR #331, opened 2020-03-26
print python coverage results to console

mikaelarguedas:
Related to https://github.com/colcon/colcon-mixin-repository/pull/21

no print to stderr
```
root@434b2f50580b:~/ws# colcon test --packages-select sros2 --pytest-with-coverage --pytest-args --cov-report=term
Starting >>> sros2   
Finished <<< sros2 [2.84s]          

Summary: 1 package finished [3.04s]

```

Example output in stdout:

```
----------- coverage: platform linux, python 3.8.2-final-0 -----------
Name                               Stmts   Miss Branch BrPart  Cover
--------------------------------------------------------------------
setup.py                              13     13      4      0     0%
sros2/__init__.py                      9      2      4      1    62%
sros2/api/__init__.py                271     58     60      8    73%
sros2/command/__init__.py              0      0      0      0   100%
sros2/command/security.py             12      2      2      1    79%
sros2/policy/__init__.py              39      5      2      1    85%
sros2/verb/__init__.py                12      2      0      0    83%
sros2/verb/create_key.py              15      3      0      0    80%
sros2/verb/create_keystore.py         14      3      0      0    79%
sros2/verb/create_permission.py       25     25      0      0     0%
sros2/verb/distribute_key.py          16     16      0      0     0%
sros2/verb/generate_artifacts.py      25     25      0      0     0%
sros2/verb/generate_policy.py         86     11     28      7    82%
sros2/verb/list_keys.py               19      3      2      1    81%
--------------------------------------------------------------------
TOTAL                                556    168    102     19    67%
Coverage HTML written to dir /root/ws/build/sros2/coverage.html
Coverage XML written to file /root/ws/build/sros2/coverage.xml

```

It can be less verbose by not printing all files 100% covered. It can also be made more verbose by printing line numbers of lines not covered

### DIFF
--- a/colcon_core/task/python/test/pytest.py
+++ b/colcon_core/task/python/test/pytest.py
@@ -92,6 +92,7 @@ class PytestPythonTestingStep(PythonTestingStepExtensionPoint):
                     '--cov-report=xml:' + str(PurePosixPath(
                         *(Path(context.args.build_base).parts)) /
                         'coverage.xml'),
+                    '--cov-report=term',
                 ]
                 # use --cov-branch option only when available
                 # https://github.com/pytest-dev/pytest-cov/blob/v2.5.0/CHANGELOG.rst
--- a/colcon_core/task/python/test/pytest.py
+++ b/colcon_core/task/python/test/pytest.py
@@ -86,13 +86,13 @@ class PytestPythonTestingStep(PythonTestingStepExtensionPoint):
                 args += [
                     '--cov=' + str(PurePosixPath(
                         *(Path(context.args.path).parts))),
+                    '--cov-report=term',
                     '--cov-report=html:' + str(PurePosixPath(
                         *(Path(context.args.build_base).parts)) /
                         'coverage.html'),
                     '--cov-report=xml:' + str(PurePosixPath(
                         *(Path(context.args.build_base).parts)) /
                         'coverage.xml'),
-                    '--cov-report=term',
                 ]
                 # use --cov-branch option only when available
                 # https://github.com/pytest-dev/pytest-cov/blob/v2.5.0/CHANGELOG.rst


dirk-thomas:
> It can be less verbose by not printing all files 100% covered. It can also be made more verbose by printing line numbers of lines not covered

I think the current state is a good default behavior.

dirk-thomas:
Thanks for the contribution.

dirk-thomas:
@mikaelarguedas Having used this change in my daily workflow since it was merged I am not so sure anymore that enabling the output by default is such a good idea. Multiple time I ran specific tests in a package to iterate on the test result and the printed coverage information was kind of in the way. Also there was no way to avoid it being printed either.

So I am leaning towards removing the default again and instead providing a different way for users to opt-in to this.

mikaelarguedas:
> Multiple times I ran specific tests in a package to iterate on the test result and the printed coverage information was kind of in the way. Also there was no way to avoid it being printed either.

Oh interesting. All the python packages I've been testing did not declare an explicit dependency on `pytest-cov` so they would not do any coverage by default and only when asked to (via pytest-args or `--pytest-with-coverage`).
It looks like it is the case for every ROS 2 python package as well.

> So I am leaning towards removing the default again and instead providing a different way for users to opt-in to this (actually it should already be possible to pass `--pytest-args --cov-report=term`).

That's fine by me, as long as there is an easy way to see coverage improvement locally.

Yes it can be done using `pytest-args` :+1:
Thanks for reopening the PR on the mixin https://github.com/colcon/colcon-mixin-repository/pull/22

dirk-thomas:
> All the python packages I've been testing did not declare an explicit dependency on pytest-cov so they would not do any coverage by default and only when asked to

Indeed, I ran into this with e.g. `colcon-core` itself: https://github.com/colcon/colcon-core/blob/52f3893c4d370384b9fc8d58f5f2dd0ca8ab2f52/setup.cfg#L55

PR #332, opened 2020-03-31
add unwrap_default_value function

dirk-thomas:
Follow up of #288 providing the inverse function.

### DIFF
--- a/colcon_core/argument_default.py
+++ b/colcon_core/argument_default.py
@@ -54,6 +54,23 @@ def wrap_default_value(value):
     return value
 
 
+def unwrap_default_value(value):
+    """
+    Unwrap a default value to the native type.
+
+    :param value: The wrapped default value
+    :returns: The unwrapped value
+    :raises ValueError: if the value is not a wrapped default value
+    """
+    global _types
+    if not is_default_value(value):
+        raise ValueError(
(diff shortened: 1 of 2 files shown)


colcon_core/argument_default.py
```diff
@@ -54,6 +54,23 @@ def wrap_default_value(value):
...
+    :param value: The wrapped default value
+    :returns: The unwrapped value
+    :raises ValueError: if the value is not a wrapped default value
+    """
+    global _types
+    if not is_default_value(value):
+        raise ValueError(
+            'the passed value is not a wrapped default value: ' +
+            str(type(value)))
+    type_ = [k for k, v in _types.items() if v == type(value)][0]
```
jacobperron: Why do we need to use list comprehension instead of something like the following?

```python
return _type[type(value)](value)
```

dirk-thomas: The comprehension is getting the *key* for `typ(value)`. Afaik there is no way to retrieve that from a `dict` without iterating it.

jacobperron: ah, I missed that it was getting the key.

dirk-thomas:
And :100:% test coverage on `argument_default.py`: :ship:

PR #333, opened 2020-03-31
Revert "print python coverage results to console"

dirk-thomas:
Reverts colcon/colcon-core#331 - see https://github.com/colcon/colcon-core/pull/331#issuecomment-606892768.


dirk-thomas:
For convenience I readded the value to the mixin: colcon/colcon-mixin-repository#22 (as originally proposed).

dirk-thomas:
@mikaelarguedas FYI.

PR #336, opened 2020-04-09
make import locally to avoid failures when not even being used

dirk-thomas:
E.g. https://build.osrfoundation.org/job/ros_gazebo_pkgs-ci-default_kinetic-xenial-amd64/110/ fails with:

```
ERROR:colcon.colcon_core.entry_point:Exception loading extension 'colcon_core.package_identification.ros': No module named 'setuptools.config', 'setuptools' needs to be at least version 30.3.0, if a newer version is not available from the package manager use 'pip3 install -U setuptools' to update to the latest version
  File "/usr/lib/python3/dist-packages/colcon_core/entry_point.py", line 101, in load_entry_points
    extension_type = load_entry_point(entry_point)
  File "/usr/lib/python3/dist-packages/colcon_core/entry_point.py", line 143, in load_entry_point
    return entry_point.load()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2229, in load
    return self.resolve()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2235, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/lib/python3/dist-packages/colcon_ros/package_identification/ros.py", line 12, in <module>
    from colcon_core.package_identification.python import get_configuration
  File "/usr/lib/python3/dist-packages/colcon_core/package_identification/python.py", line 12, in <module>
    from setuptools.config import read_configuration
ImportError: No module named 'setuptools.config', 'setuptools' needs to be at least version 30.3.0, if a newer version is not available from the package manager use 'pip3 install -U setuptools' to update to the latest version
```

even though no packages in the workspace are actually Python packages.

This patch moves the import into local scope to only happen when actually being used.

### DIFF
--- a/colcon_core/package_identification/python.py
+++ b/colcon_core/package_identification/python.py
@@ -8,20 +8,6 @@ from colcon_core.package_identification \
 from colcon_core.plugin_system import satisfies_version
 from distlib.util import parse_requirement
 from distlib.version import NormalizedVersion
-try:
-    from setuptools.config import read_configuration
-except ImportError as e:
-    from pkg_resources import get_distribution
-    from pkg_resources import parse_version
-    setuptools_version = get_distribution('setuptools').version
-    minimum_version = '30.3.0'
-    if parse_version(setuptools_version) < parse_version(minimum_version):
-        e.msg += ', ' \
-            "'setuptools' needs to be at least version {minimum_version}, if" \
-            ' a newer version is not available from the package manager use ' \
-            "'pip3 install -U setuptools' to update to the latest version" \
-            .format_map(locals())
-    raise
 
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
> So `colcon_ros` will still import that function, but this import won't happen unless `colcon_ros` actually calls the function, right?
> So we shouldn't see the message unless we encounter an `ament_python` package and we don't have an adequate `setuptools` installed.

Correct: https://github.com/colcon/colcon-ros/blob/94856c2f302e8e1df685d4faf70b3856fb6963d6/colcon_ros/package_identification/ros.py#L109-L115

PR #348, opened 2020-05-08
[windows] fix _colcon_python_executable fallback code path.

seanyen:
In the `local_setup.bat`, it first attempt to use the Python binary decided at compile time. If it doesn't exist, it will then test the first one visible from `%PATH%`. In the cases of no Python in the `%PATH%`, it will error on with `error: unable to find python executable`.

However, in my setting, I don't have any Python binary in `%PATH%`, but I do have the Python binary under `@(python_executable)` which decided at the compile time. By the logic, it should firstly find that one, but it error out with `error: unable to find python executable` instead.

After digging, it turned out that when accessing the variable is set within a FOR loop or IF block, one have to turn on `setlocal enabledelayedexpansion` and use `!_colcon_python_executable!` to access the values. This pull request is to correct this usage.

### DIFF
--- a/colcon_core/shell/template/prefix.bat.em
+++ b/colcon_core/shell/template/prefix.bat.em
@@ -52,6 +52,7 @@ goto:eof
 :: Run the commands in topological order
 :: first argument: the base path to look for packages
 :_colcon_run_ordered_commands
+  setlocal enabledelayedexpansion
   :: check environment variable for custom Python executable
   if "%COLCON_PYTHON_EXECUTABLE%" NEQ "" (
     if not exist "%COLCON_PYTHON_EXECUTABLE%" (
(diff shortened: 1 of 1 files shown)


seanyen:
@dirk-thomas This is ready for review and merge. Thanks!

dirk-thomas:
Thanks for the patch.

seanyen:
@dirk-thomas Just a quick question, do you know when we might be able to have another release for `0.5.10`?

dirk-thomas:
> do you know when we might be able to have another release for 0.5.10?

The Foxy release preparations have kept me pretty busy in the past weeks. I hope I will get a new release out within a week.

(You can enable notifications for "Releases only" to get an automated ping when that happens.)

dirk-thomas:
:ship: 0.5.10.

seanyen:
Hooray!

PR #367, opened 2020-08-07
Bugfix missing explicit key updates in update_descriptor

stonier:
The implementation for that described in the method comment was missing, leading to this issue for colcon_meta &rarr; https://github.com/colcon/colcon-metadata/issues/23.

### DIFF
--- a/colcon_core/package_augmentation/__init__.py
+++ b/colcon_core/package_augmentation/__init__.py
@@ -135,6 +135,15 @@ def update_descriptor(
     :param additional_argument_names: A dict of option names to destination
       names or a list of argument names
     """
+    # explicit key updates
+    try:
+        desc.name = data['name']
+    except KeyError:
+        pass
+    try:
+        desc.type = data['type']
+    except KeyError:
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Please see https://github.com/colcon/colcon-metadata/issues/23#issuecomment-670709086 why altering package identification information during the augmentation are phase isn't supported.

stonier:
What are the ramifications? I haven't been able to 'break' things (`colcon graph` and co. all work fine). 

Probably, the comments just above the code change should be edited if this is true, i.e. it currently states:

```
For the keys `name` and `type` the values from `data` are set on the
descriptor attributes with the same names.
```

dirk-thomas:
> What are the ramifications?

Another extension might have already contributed information on the previous package name which would potentially be incorrect. The augmentation phase is simply to late to change that identifying piece of a package. It can only be set during discovery / identification. Please see colcon/colcon-metadata#24 how to specify the package name in the `paths` section of `.meta` file.

PR #386, opened 2020-09-24
support renaming through package_dir

### LINKED ISSUE #272: Python packages built with --symlink-install does not respect packages_dir
Follow up from #271

[`package_dir={'foo': 'lib'}` means `lib/__init__.py` is installed as `foo/__init__.py`](https://docs.python.org/3.8/distutils/setupscript.html#listing-whole-packages), however that currently doesn't work when using `--symlink-install`. This is pretty much the same as  pypa/setuptools#230 which says `python setup.py develop` doesn't do the right thing either. However, since `colcon-core` isn't using the `.pth` normally created by `setup.py develop` it does seem fair to call it a difference between `colcon build` and `colcon build --symlink-install`.

To reproduce

```console
sloretz@nxt:~/ws/misc-py/src$ tree bob
bob
├── lib
│   └── __init__.py
└── setup.py

1 directory, 2 files
sloretz@nxt:~/ws/misc-py/src$ cat bob/lib/__init__.py 
print('hello bob')
sloretz@nxt:~/ws/misc-py/src$ cat bob/setup.py 
from setuptools import setup

setup(
    name='bob',
    version='0.1.0',
    packages=['bob'],
    package_dir={'bob': 'lib'},
    install_requires=['setuptools'],
)
```

Run `colcon build` and in python `import bob`. It works as expected.

```
sloretz@nxt:~/ws/misc-py$ . install/setup.bash 
sloretz@nxt:~/ws/misc-py$ python3
Python 3.6.8 (default, Oct  7 2019, 12:59:55) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import bob
hello bob
```

Then run clean the workspace, run `colcon build --symlink-install` and try importing `bob`, it doesn't work, but importing `lib` does.

```
sloretz@nxt:~/ws/misc-py$ . install/setup.bash 
sloretz@nxt:~/ws/misc-py$ python3
Python 3.6.8 (default, Oct  7 2019, 12:59:55) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import bob
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'bob'
>>> import lib
hello bob
```

dirk-thomas:
Fixes #272.

### DIFF
--- a/colcon_core/task/python/build.py
+++ b/colcon_core/task/python/build.py
@@ -186,16 +186,31 @@ class PythonBuildTask(TaskExtensionPoint):
 
     def _symlinks_in_build(self, args, setup_py_data):
         items = ['setup.py']
+        renamed_items = []
         # add setup.cfg if available
         if os.path.exists(os.path.join(args.path, 'setup.cfg')):
             items.append('setup.cfg')
         # add all first level packages
         package_dir = setup_py_data.get('package_dir') or {}
-        for package in setup_py_data.get('packages') or []:
+        packages = setup_py_data.get('packages') or []
+        for package in packages:
             if '.' in package:
                 continue
             if package in package_dir:
                 items.append(package_dir[package])
+                renamed_items.append((package_dir[package], package))
+                if package_dir[package] in package_dir:
+                    package_dir_package = package_dir[package]
+                    raise RuntimeError(
+                        "The package_dir contains a mapping from '{package}' "
+                        "to '{package_dir_package}' which is also a key"
+                        .format_map(locals()))
+                if package_dir[package] in packages:
+                    package_dir_package = package_dir[package]
+                    raise RuntimeError(
+                        "The value '{package_dir_package}' in package_dir is "
+                        'also listed in packages'
+                        .format_map(locals()))
             elif '' in package_dir:
                 items.append(os.path.join(package_dir[''], package))
             else:
@@ -220,10 +235,20 @@ class PythonBuildTask(TaskExtensionPoint):
                 source = os.path.relpath(source, args.path)
             items.append(source)
 
+        symlinks = []
         # symlink files / directories from source space into build space
         for item in items:
-            src = os.path.join(args.path, item)
-            dst = os.path.join(args.build_base, item)
+            symlinks.append((
+                os.path.join(args.path, item),
+                os.path.join(args.build_base, item)))
+        # provide a symlink within the build space if a module name is
+        # changed by the mapping specified in package_dir
+        for rel_src, rel_dst in renamed_items:
+            symlinks.append((
+                os.path.join(args.build_base, rel_src),
+                os.path.join(args.build_base, rel_dst)))
+
+        for src, dst in symlinks:
             os.makedirs(os.path.dirname(dst), exist_ok=True)
             if os.path.islink(dst):
                 if not os.path.exists(dst) or not os.path.samefile(src, dst):


colcon_core/task/python/build.py
```diff
@@ -220,10 +235,20 @@ def _symlinks_in_build(self, args, setup_py_data):
...
             items.append(source)
 
+        symlinks = []
         # symlink files / directories from source space into build space
         for item in items:
-            src = os.path.join(args.path, item)
-            dst = os.path.join(args.build_base, item)
+            symlinks.append((
+                os.path.join(args.path, item),
+                os.path.join(args.build_base, item)))
```
sloretz, line 243: Since the base baths are still symlinked, in the bob case this causes it to still be importable as `lib`

```
>>> import bob
Hello from bob
>>> import lib
Hello from bob
```

colcon_core/task/python/build.py
```diff
@@ -186,16 +186,31 @@ def _undo_install(self, pkg, args, setup_py_data, python_lib):
...
+                renamed_items.append((package_dir[package], package))
+                if package_dir[package] in package_dir:
+                    package_dir_package = package_dir[package]
+                    raise RuntimeError(
+                        "The package_dir contains a mapping from '{package}' "
+                        "to '{package_dir_package}' which is also a key"
+                        .format_map(locals()))
+                if package_dir[package] in packages:
+                    package_dir_package = package_dir[package]
+                    raise RuntimeError(
```
sloretz, line 210: This definitely improves the support for `package_dir`. It doesn't support the following case, but I'm not sure the following case is worth supporting. Raising an error here is definitely better than silently doing the wrong thing like `python setup.py develop` does.

```
alice
├── dee
│   └── __init__.py
├── dum
│   └── __init__.py
└── setup.py
```

```python
from setuptools import setup

setup(
    name='alice',
    version='0.1.0',
    packages=['dee', 'dum'],
    package_dir={'dee': 'dum', 'dum': 'dee'},
    install_requires=['setuptools'],
)
```

With `python setup.py install` the two modules are swapped

```
>>> import dee
originally dum
>>> import dum
originally dee
```

With `python setup.py develop` the modules are incorrectly not swapped, and no error

```
>>> import dee
originally dee
>>> import dum
originally dum
```

with `colcon build --symlink-install` it raises an error about this case not being supported.

```
  File "/tmp/colcon_pr_test/colcon_ws/colcon-from-source/install/colcon-core/lib/python3.6/site-packages/colcon_core/executor/__init__.py", line 91, in __call__
    rc = await self.task(*args, **kwargs)
  File "/tmp/colcon_pr_test/colcon_ws/colcon-from-source/install/colcon-core/lib/python3.6/site-packages/colcon_core/task/__init__.py", line 93, in __call__
    return await task_method(*args, **kwargs)
  File "/tmp/colcon_pr_test/colcon_ws/colcon-from-source/install/colcon-core/lib/python3.6/site-packages/colcon_core/task/python/build.py", line 88, in build
    self._symlinks_in_build(args, setup_py_data)
  File "/tmp/colcon_pr_test/colcon_ws/colcon-from-source/install/colcon-core/lib/python3.6/site-packages/colcon_core/task/python/build.py", line 207, in _symlinks_in_build
    .format_map(locals()))
RuntimeError: The package_dir contains a mapping from 'dee' to 'dum' which is also a key
```

sloretz (approved):
Looks like an improvement to me!

colcon_core/task/python/build.py
dirk-thomas: Correct. Since the directory `lib` needs to be preserved so the `setup.py` file can work as intended there isn't anything we can do about it (afaik).

colcon_core/task/python/build.py
dirk-thomas: > it raises an error about this case not being supported.

Yeah, this is the best we can do (afaik).

colcon_core/task/python/build.py
sloretz: Random thought - can the symlink to the base path be removed after invoking `setup.py`?

dirk-thomas: > Can the symlink to the base path be removed after invoking `setup.py`?

None of the symlinks are removed at the moment. Maybe it could be. Is there a specific reason why it should be?

sloretz: > Maybe it could be. Is there a specific reason why it should be?

I was wondering if this symlink was removed afterwards then maybe `import lib` would no longer also import bob.

dirk-thomas: Could you give that a try? If that works that would be nicer / cleaner.

sloretz: Opened proof of concept in https://github.com/colcon/colcon-core/pull/390

PR #387, opened 2020-09-25
Pass --egg-base as relative.

mikepurvis:
It seems setuptools doesn't like `--egg-base` to be an absolute path. Example with setuptools 45.2.0 (Ubuntu Focal):

```
--- stderr: canopen
error: Error: setup script specifies an absolute path:

    /home/administrator/ws/build/canopen/canopen.egg-info/PKG-INFO

setup() arguments must *always* be /-separated paths relative to the
setup.py directory, *never* absolute paths.

---
Failed   <<< canopen [1.41s, exited with code 1]
```

The error goes away if I compute the `--egg-base` argument as a path relative to the cwd, as in this PR.

### DIFF
--- a/colcon_core/task/python/build.py
+++ b/colcon_core/task/python/build.py
@@ -67,9 +67,10 @@ class PythonBuildTask(TaskExtensionPoint):
 
             # invoke `setup.py install` step with lots of arguments
             # to avoid placing any files in the source space
+            rel_build_base = os.path.relpath(args.build_base, args.path)
             cmd = [
                 executable, 'setup.py',
-                'egg_info', '--egg-base', args.build_base,
+                'egg_info', '--egg-base', rel_build_base,
                 'build', '--build-base', os.path.join(
                     args.build_base, 'build'),
                 'install', '--prefix', args.install_base,
--- a/colcon_core/task/python/build.py
+++ b/colcon_core/task/python/build.py
@@ -67,10 +67,10 @@ class PythonBuildTask(TaskExtensionPoint):
 
             # invoke `setup.py install` step with lots of arguments
             # to avoid placing any files in the source space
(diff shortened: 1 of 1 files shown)


mikepurvis:
Slightly related, Python packages which use `distutils.core.setup` don't support the `egg_base` parameter, example:

```
Starting >>> python-npyscreen
--- stderr: python-npyscreen
   or: setup.py --help-commands
   or: setup.py cmd --help

error: invalid command 'egg_info'
---
Failed   <<< python-npyscreen [0.37s, exited with code 1]
```

I've hacked around this for now by just patching that package to use `setuptools` instead, but this may be worth attempting to detect, depending how committed colcon is to supporting a broad array of pypi packages.

dirk-thomas:
Can you please provide reproducible since the problem doesn't happen for me with usual invocations.

mikepurvis:
In case there's any doubt, here it is with a very popular package which should definitely just work:

```
python3 -m venv colcon-env
colcon-env/bin/pip install colcon-common-extensions
mkdir ws; cd ws
git clone https://github.com/psf/requests.git
~/colcon-env/bin/colcon build
```

Result:

```
Starting >>> requests
--- stderr: requests
warning: no files found matching 'requirements.txt'
error: Error: setup script specifies an absolute path:

    /home/administrator/ws2/build/requests/requests.egg-info/PKG-INFO

setup() arguments must *always* be /-separated paths relative to the
setup.py directory, *never* absolute paths.

---
Failed   <<< requests [0.87s, exited with code 1]

Summary: 0 packages finished [1.74s]
  1 package failed: requests
  1 package had stderr output: requests
```

dirk-thomas:
Thanks for the patch. I just inlined the single use local variable in 0c757be2144a009fed1ad1e1bcde875cb9b6bf01.

mikepurvis:
LGTM

PR #390, opened 2020-09-25
Remove unnecessary symlinks after building modules renamed by packages_dir

sloretz:
Idea from https://github.com/colcon/colcon-core/pull/386/files#r495217794

Some symlinks are needed by setup.py for packages renamed via
`packages_dir`, but can be discarded after. Discarding them makes sure the
renamed modules aren't importable by their unrenamed name.

Without this PR
```
>> import bob
Hello from bob
>>> import lib
Hello from bob
```

With this PR

```
>>> import bob
Hello from bob
>>> import lib
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'lib'
```

### DIFF
--- a/colcon_core/task/python/build.py
+++ b/colcon_core/task/python/build.py
@@ -7,7 +7,6 @@ with suppress(ImportError):
     # to avoid warning introduced in setuptools 49.2.0
     import setuptools  # noqa: F401
 from distutils.sysconfig import get_python_lib
-import functools
 import locale
 import os
 from pathlib import Path
@@ -95,7 +94,7 @@ class PythonBuildTask(TaskExtensionPoint):
 
         else:
             self._undo_install(pkg, args, setup_py_data, python_lib)
-            unlink_temp_symlinks = self._symlinks_in_build(args, setup_py_data)
+            temp_symlinks = self._symlinks_in_build(args, setup_py_data)
 
             # invoke `setup.py develop` step in build space
             # to avoid placing any files in the source space
@@ -117,7 +116,9 @@ class PythonBuildTask(TaskExtensionPoint):
                 completed = await run(
                     self.context, cmd, cwd=args.build_base, env=env)
             finally:
-                unlink_temp_symlinks()
+                # Remove symlinks that were only needed during build time
+                for symlink in temp_symlinks:
+                    os.unlink(symlink)
 
             if completed.returncode:
                 return completed.returncode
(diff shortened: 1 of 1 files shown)


colcon_core/task/python/build.py
```diff
@@ -244,10 +249,14 @@ def _symlinks_in_build(self, args, setup_py_data):
                 os.path.join(args.build_base, item)))
         # provide a symlink within the build space if a module name is
         # changed by the mapping specified in package_dir
+        temp_symlinks = []
         for rel_src, rel_dst in renamed_items:
             symlinks.append((
-                os.path.join(args.build_base, rel_src),
+                os.path.join(args.path, item),
```
sloretz, line 255: this makes the symlink point to source space because the path it's pointing to gets deleted by the other new code in this PR

dirk-thomas, line 255: `item` seems wrong here since it is not set in this loop but only by earlier code.

colcon_core/task/python/build.py
```diff
@@ -105,6 +105,11 @@ async def build(self, *, additional_hooks=None):  # noqa: D102
             self._append_install_layout(args, cmd)
             completed = await run(
                 self.context, cmd, cwd=args.build_base, env=env)
+
+            # Remove symlinks that were only needed during build time
```
dirk-thomas, line 109: Should this happen in a `finally` block instead to ensure it is done even if the in-between code raises an exception?

dirk-thomas:
@sloretz Friendly ping.

colcon_core/task/python/build.py
sloretz: Fixed, but accidentally squashed fix during a rebase. Latest state does it in a finally block.

colcon_core/task/python/build.py
sloretz: Oops, `item` -> `rel_src` in d425ef1. This time I tested locally with a package that has two renamed items.

colcon_core/task/python/build.py
```diff
@@ -289,6 +299,13 @@ def _symlinks_in_build(self, args, setup_py_data):
             if not os.path.exists(dst):
                 os.symlink(src, dst)
 
+        def unlink_symlinks(symlinks):
+            # Remove symlinks that were only needed during build time
+            for symlink in symlinks:
+                os.unlink(symlink)
+
+        return functools.partial(unlink_symlinks, temp_symlinks)
```
dirk-thomas, line 307: I would prefer to keep this simple and just return `temp_symlinks` here and above in the `finally` block loop over the returned symlinks and `unlink()` them.

sloretz: Return iterable and unlink in `finally` instead of returning callable in f5b8bae

dirk-thomas (approved):
Thanks for this improvement.

colcon_core/task/python/build.py
mikepurvis: Overkill in this instance, but another way to defer cleanup like this can be to `ensure_future` on a coroutine that just sits there waiting to be canceled and then performs the cleanup work on cancellation. I've used this approach to do things like clean up a shared database connection when an async webserver exits.

PR #396, opened 2020-10-05
support building Python packages using distutils

### LINKED ISSUE #394: Not all python packages support egg_base
A small leftover from https://github.com/colcon/colcon-core/pull/387#issuecomment-698937014 is that some Python packages from PyPI don't support the `egg_base` option. We can work on getting fixes push upstream to these (https://github.com/msoulier/tftpy/pull/114, https://github.com/npcole/npyscreen/pull/106), but it would be nice if colcon could detect this failure and retry without that option.

Reproduction:

```
git clone https://github.com/npcole/npyscreen
colcon build
```

Result:

```
Starting >>> python3-npyscreen
--- stderr: python3-npyscreen
   or: setup.py --help-commands
   or: setup.py cmd --help

error: invalid command 'egg_info'
---
Failed   <<< python3-npyscreen [0.72s, exited with code 1]
```

We've found a handful of python packages in our distribution which fail in this way.

dirk-thomas:
Fixes #394.

### DIFF
--- a/colcon_core/task/python/build.py
+++ b/colcon_core/task/python/build.py
@@ -7,6 +7,7 @@ with suppress(ImportError):
     # to avoid warning introduced in setuptools 49.2.0
     import setuptools  # noqa: F401
 from distutils.sysconfig import get_python_lib
+import locale
 import os
 from pathlib import Path
 import shutil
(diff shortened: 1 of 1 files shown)


mikepurvis:
This works for my use case, thanks for looking at it. :+1:

gbiggs (approved):

mikepurvis:
FWIW, I recall encountering a PyPI package which didn't even support `--help-commands`, but I have no idea what it was, possibly a super-old version of pyyaml? Anyway, I don't think that's worth thinking too much about in 2020— certainly with this change you probably have like 99% of coverage of all the stuff which actually matters.

dirk-thomas:
If we run into another packages which doesn't work with this we can look into how to make it work.

PR #405, opened 2020-11-04
fix bootstrapping not using new Python augmentation extension

dirk-thomas:
The regression was introduced in #385 when some functionality was extracted from the Python identification extension into a new Python augmentation extension. This patch makes sure that new extension is available during bootstrapping.


gbiggs (approved):

gbiggs:
LGTM, tested locally and works.

PR #434, opened 2021-05-13
Consider alternate pytest_cov version string format

### LINKED ISSUE #433: Branch coverage information can't be generated with pytest-cov>=2.10.1
Reproduce by testing a Python package and using the `--pytest-with-coverage` option (or `--mixin coverage-pytest`) with `pytest-cov>=2.10.1`:

```
$ colcon test --packages-select tracetools_test --mixin coverage-pytest 
Starting >>> tracetools_test
[3.992s] WARNING:colcon.colcon_core.verb.test:Test coverage will be produced but will not contain branch coverage information because the pytest extension 'cov' does not support it (need 2.5.0, have __version__ = '2.11.1')
--- stderr: tracetools_test                   
Coverage.py warning: No data was collected. (no-data-collected)
---
Finished <<< tracetools_test [6.59s]    [ with test failures ]
```

This is because `pytest_cov.__version__ == "__version__ = '2.10.1'"` (instead of `pytest_cov.__version__ == '2.10.1'`) since 2.10.1: https://github.com/pytest-dev/pytest-cov/commit/da61b912c48fadf6df71c8cd4712a23dbdc2efc0. I really have no idea why, but I opened an issue upstream: https://github.com/pytest-dev/pytest-cov/issues/468

This breaks the version condition here: https://github.com/colcon/colcon-core/blob/e775de7fb30060f4a220d6754f68e7940f53d85a/colcon_core/task/python/test/pytest.py#L98

Example:

```py
>>> from pkg_resources import parse_version
>>> from pytest_cov import __version__ as pytest_cov_version
>>> pytest_cov_version
"__version__ = '2.11.1'"
>>> parse_version('2.11.1') >= parse_version('2.5.0')
True
>>> parse_version(pytest_cov_version) >= parse_version('2.5.0')
False
```

Thus branch coverage information is never enabled.

Adding a workaround should be easy, though.

christophebedard:
Fixes #433

Try to check with the "new" version string pattern to see if there is a match. If there is one, use that as the `x.y.z` version.

Signed-off-by: Christophe Bedard <bedard.christophe@gmail.com>

### DIFF
--- a/colcon_core/task/python/test/pytest.py
+++ b/colcon_core/task/python/test/pytest.py
@@ -4,6 +4,7 @@
 import os
 from pathlib import Path
 from pathlib import PurePosixPath
+import re
 import sys
 
 from colcon_core.event.test import TestFailure
@@ -77,6 +78,13 @@ class PytestPythonTestingStep(PythonTestingStepExtensionPoint):
         ):
             try:
                 from pytest_cov import __version__ as pytest_cov_version
+                # version string might not be simply 'x.y.z' for some versions
+                # https://github.com/pytest-dev/pytest-cov/issues/468
+                alt_version_pattern = re.compile(
+                    r"__version__ = '([0-9]+\.[0-9]+\.[0-9]+)'")
+                alt_version = alt_version_pattern.findall(pytest_cov_version)
+                if alt_version:
+                    pytest_cov_version = alt_version[0]
             except ImportError:
                 logger.warning(
                     'Test coverage will not be produced for package '


christophebedard:
This works for me. I'm wondering if this is going to be considered a bug (https://github.com/pytest-dev/pytest-cov/issues/468). In any case, this fix is needed for `pytest_cov=={2.10.1,2.11.0,2.11.1}`.

christophebedard:
> This works for me. I'm wondering if this is going to be considered a bug ([pytest-dev/pytest-cov#468](https://github.com/pytest-dev/pytest-cov/issues/468)). In any case, this fix is needed for `pytest_cov=={2.10.1,2.11.0,2.11.1}`.

It was considered a bug and fixed: https://github.com/pytest-dev/pytest-cov/commit/f5a4d7e008087913ce2bc9dc2305d05a5f545a41. So it should be fine for `>2.11.1`, but for `2.10.1` to `2.11.1` we need this PR.

dirk-thomas:
Since a new version of `pytest-cov` was released I don't see a case where anyone would install a version from the range where it was broken. Do you agree that this can be closed?

christophebedard:
> Since a new version of `pytest-cov` was released I don't see a case where anyone would install a version from the range where it was broken. Do you agree that this can be closed?

Yeah sure, that's fine.

dirk-thomas:
Thanks for your effort to get this resolved upstream!

PR #444, opened 2021-09-17
Use extras_require[test] instead of tests_require

sloretz:
This replaces `tests_require` with a pattern that seems to have popped
up since pypa/setuptools#931 and pypa/setuptools#1684. It allows test
dependencies to be installed by adding `[test]` to the end of the
package name.

Example installing test dependencies and running pytest:

```
python3 -m venv env3
. env3/bin/activate
cd colcon-core/
pip install -e .[test]
pytest
```

I chose `test`, but the naming is inconsistent. I've found other
examples using `tests` or `testing` as well.

I opened this because it could get rid of the duplication of test dependencies from the CI scripts in #414 (@cottsay FYI)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,7 +38,10 @@ install_requires =
   pytest-rerunfailures
   setuptools>=30.3.0
 packages = find:
-tests_require =
+zip_safe = false
+
+[options.extras_require]
+test =
   flake8>=3.6.0
   flake8-blind-except
   flake8-builtins
@@ -54,7 +57,6 @@ tests_require =
   pytest
   pytest-cov
   scspell3k>=2.2
-zip_safe = false
 
 [options.packages.find]
 exclude =


mikepurvis:
Thanks for looking into this. I've been desperately frustrated that there isn't more consistency in the Python world for how to manage different classes of dependencies— testing, linting, mypy stubs, etc.

When I looked a few months ago, the closest thing seemed to be PEP 0631:

https://www.python.org/dev/peps/pep-0631/#specification

But the whole thing still feels underspecified and undersupported by the tooling. It's a nightmare for someone wanting to stand up generic CI scripts that just figure out what to do. Anyway, using `options.extras_require.test` seems like a reasonable stepping stone to the eventual `project.optional-dependencies.tests` suggested for use in `pyproject.toml` by the PEP.

cottsay (approved):
This seems to work flawlessly. Thanks!

setup.cfg
```diff
@@ -38,24 +38,11 @@ install_requires =
...
-  mock
-  pep8-naming
-  pylint
-  pytest
-  pytest-cov
-  scspell3k>=2.2
 zip_safe = false
 
+[options.extras_require]
+test = flake8>=3.6.0; flake8-blind-except; flake8-builtins; flake8-class-newline; flake8-comprehensions; flake8-deprecated; flake8-docstrings; flake8-import-order; flake8-quotes; mock; pep8-naming; pylint; pytest; pytest-cov; scspell3k>=2.2
```
nuclearsandwich, line 44: Was there a specific motivation for the oneline approach versus something that matches the previous tests_require stanza?

```ini
[options.extras_require]
test =
 flake8>=3.6.0
 flake8-blind-except
 flake8-builtins
 flake8-class-newline
 flake8-comprehensions
...
```

seems easier to read and edit to me.

sloretz: > specific motivation

Blind following of a stack overflow post without much thought :no_mouth:

sloretz: Using newlines in 3cdb542

nuclearsandwich (approved):

sloretz:
@dirk-thomas would you like to review this one, or may I merge it?

dirk-thomas:
If it works for you, please feel free to merge.

(Some CI seems to fail though.)

sloretz:
> (Some CI seems to fail though.)

Closing and reopening to see if that retriggers CI. I can't think of a reason the `test_create_event_reactor` test would fail given the changes, so I'm wondering if it's flaky.

PR #449, opened 2021-10-07
Warn when overriding packages at build time

sloretz:
Part of https://github.com/ros2/ros2/issues/1150

This PR adds a new warning to `colcon build` when a package to be built would override one in an underlay workspace. It also adds an option `--alow-overriding` to disable the a warning and allow building it anyways.

## Trying this PR out
1. Bootstrap colcon from source with this PR and the companion PR in colcon/colcon-ros#129
2. Install `ros-rolling-desktop`
3. Create a colcon workspace and one or more ROS 2 repos in it (below I did `git@github.com:ros2/rcutils.git` and `git@github.com:ros2/rcl.git`)
4. source `/opt/ros/rolling/setup.bash`
4. Run `colcon build`

```console
$ colcon build
[0.387s] ERROR:colcon:colcon build: Some selected packages are already built in one or more underlay workspaces:
	'rcutils' is in: /opt/ros/rolling
	'rcl_yaml_param_parser' is in: /opt/ros/rolling
	'rcl' is in: /opt/ros/rolling
	'rcl_action' is in: /opt/ros/rolling
	'rcl_lifecycle' is in: /opt/ros/rolling
If a package in a merged underlay workspace is overridden and it installs headers, then all packages in the overlay must sort their include directories by workspace order. Failure to do so may result in build failures or undefined behavior at run time.
If the overridden package is used by another package in any underlay, then the overriding package in the overlay must be API and ABI compatible or undefined behavior at run time may occur.

If you understand the risks and want to override a package anyways, add the following to the command line:
	--allow-overriding rcl rcl_action rcl_lifecycle rcl_yaml_param_parser rcutils
```

## How it works

This adds an extension point `FindInstalledPackagesExtensionPoint` which allows extensions to add to the packages found when `find_installed_packages(...)` is called. This hook allows `colcon-ros` to implement an extension point that reads the ament index to find packages. In the `main()` of the `Build` verb it finds all installed packages in chained prefixes, and checks if any package being built already exists in one of them. If so, the warning is emitted.

## Limitations

Warning at build time is not perfect, because it's possible a user will install a package into the underlay `/opt/ros/...` later that was built by this overlay beforehand. I tried looking at how to make the setup scripts warn when they're sourced, but it seems difficult to do cleanly in the case of `/opt/ros/...`.


colcon_core/shell/__init__.py
```diff
@@ -563,46 +562,93 @@ def find_installed_packages_in_environment():
...
-    Based on the install layout the packages are discovered i different
+    Based on the install layout the packages may be discovered in different
     locations.
 
     :param Path install_base: The base path to find installed packages in
     :returns: The mapping from a package name to the prefix path, None if the
-      path doesn't exist or doesn't contain a marker file with a valid install
-      layout
-    :rtype: OrderedDict or None
+      path doesn't exist or doesn't a valid install layout
```
jacobperron, line 616:
```suggestion
      path doesn't exist or is not a supported install layout
```

colcon_core/verb/build.py
```diff
@@ -133,6 +160,45 @@ def main(self, *, context):  # noqa: D102
         jobs, unselected_packages = self._get_jobs(
             context.args, decorators, install_base)
 
+        underlay_packages = {}
+        for prefix_path in get_chained_prefix_path():
+            packages = find_installed_packages(Path(prefix_path))
+            for pkg, path in packages.items():
```
jacobperron, line 166: We might want to guard against `packages` being `None` (it's a possible return value).

sloretz: Guards against packages being `None` in ac6cf92

colcon_core/shell/__init__.py
sloretz: Fixed grammar in 
6e6b45e

jacobperron (approved):
LGTM

sloretz:
@dirk-thomas would you like to review this and colcon/colcon-ros#129, or may I merge them?

dirk-thomas:
@sloretz since I don't have the bandwidth to review `colcon` changes in my spare time, please feel free to move forward at your own discretion.

On a very brief look the reduction in coverage might be something to consider addressing. Also the core changes will likely need to be released first, with a proper version dependency in the non-core package.

sloretz:
> On a very brief look the reduction in coverage might be something to consider addressing

I added  a couple more tests for edge cases, which I think is the maximum amount of testable code in this PR.

> Also the core changes will likely need to be released first, with a proper version dependency in the non-core package.

Noted :+1:

jacobperron (approved):
Latest changes also LGTM

colcon_core/shell/__init__.py
```diff
@@ -563,46 +562,93 @@ def find_installed_packages_in_environment():
...
-        # find all files in the subdirectory
-        if (install_base / get_relative_package_index_path()).is_dir():
-            package_index = install_base / get_relative_package_index_path()
-            for p in package_index.iterdir():
-                if not p.is_file():
-                    continue
-                if p.name.startswith('.'):
-                    continue
-                packages[p.name] = install_base
+            if pkg in packages and not path.samefile(packages[pkg]):
```
jacobperron, line 641: Is this check working as expected? I can produce a warning when a package only exists in one workspace, e.g.:

```
mkdir src
ros2 pkg create foo --destination-directory src
colcon build
source install/setup.bash
colcon build
```

Or perhaps this is not a recommended workflow.

jacobperron: Fixed by https://github.com/colcon/colcon-core/pull/473

PR #472, opened 2022-01-21
Setup find_installed_packages to fix bootstrap instructions

sloretz:
Fix an issue from #449 - the bootstrap from source instructions are failing with the following error

```
[0.267s] ERROR:colcon.colcon_core.shell:Exception in shell extension 'sh': get_entry_points() not overridden for group 'colcon_core.shell.find_installed_packages'
Traceback (most recent call last):
  File "/home/sloretz/ws/colcon-from-source/src/colcon-core/colcon_core/shell/__init__.py", line 298, in get_command_environment
    return await extension.generate_command_environment(
  File "/home/sloretz/ws/colcon-from-source/src/colcon-core/colcon_core/shell/sh.py", line 137, in generate_command_environment
    check_dependency_availability(
  File "/home/sloretz/ws/colcon-from-source/src/colcon-core/colcon_core/shell/__init__.py", line 510, in check_dependency_availability
    packages_in_env = find_installed_packages_in_environment()
  File "/home/sloretz/ws/colcon-from-source/src/colcon-core/colcon_core/shell/__init__.py", line 552, in find_installed_packages_in_environment
    pkgs = find_installed_packages(prefix_path)
  File "/home/sloretz/ws/colcon-from-source/src/colcon-core/colcon_core/shell/__init__.py", line 621, in find_installed_packages
    prioritized_extensions = get_find_installed_packages_extensions()
  File "/home/sloretz/ws/colcon-from-source/src/colcon-core/colcon_core/shell/__init__.py", line 601, in get_find_installed_packages_extensions
    extensions = instantiate_extensions(__name__ + '.find_installed_packages')
  File "/home/sloretz/ws/colcon-from-source/src/colcon-core/colcon_core/plugin_system.py", line 36, in instantiate_extensions
    extension_types = load_entry_points(
  File "./src/colcon-core/bin/colcon", line 28, in custom_load_entry_points
    assert group_name in custom_entry_points, \
AssertionError: get_entry_points() not overridden for group 'colcon_core.shell.find_installed_packages'
```

I thought CI covered the bootstrap case, so I'm not sure why this wasn't caught.

PR #473, opened 2022-01-21
Don't warn about overriding when building the a workspace that has al…

### LINKED ISSUE #465: Overriding package error for packages in current workspace
I've just experienced a slightly annoying side effect of this recently merged PR: https://github.com/colcon/colcon-core/pull/449 

Consider the following workflow:

```bash
source /opt/ros/noetic/setup.bash
cd my_workspace
colcon build --merge-install
source install/setup.bash
colcon build --merge-install
```
I use this workflow often because I build, run something, make some changes, then want to build again. However, the second colcon build fails because the workspace is now seen as overlaying itself and it prints the package-overridden error message for every package in the local workspace - it thinks they are overriding themselves.

Is my workflow bad (i.e. running colcon build after sourcing the local workspace) or is this a bug?

sloretz:
Requires #472 (currently targeting that PR's branch)
Resolves #465

If a user builds a workspace, sources the install space, then builds again the override warning was triggering. This PR avoids warning in that case. It does so by ignoring the install space of the current workspace when looking for packages in underlay's.

### DIFF
--- a/colcon_core/verb/build.py
+++ b/colcon_core/verb/build.py
@@ -162,6 +162,10 @@ class BuildVerb(VerbExtensionPoint):
 
         underlay_packages = {}
         for prefix_path in get_chained_prefix_path():
+            if prefix_path == install_base:
+                # The install space of the workspace being built is sourced
+                # Ignore it for the purposes of the override warning
+                continue
             packages = find_installed_packages(Path(prefix_path))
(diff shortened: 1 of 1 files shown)


jacobperron (approved):
LGTM. Consider adding a test case to maintain code coverage.

PR #487, opened 2022-02-19
Make output of colcon colorful using colorama

timonegk:
This pull request replaces #195. In general, it provides the same coloring, heavily inspired by catkin_tools. To do the formatting, it uses colorama, a library that also supports Windows.

### DIFF
--- a/colcon_core/event_handler/console_start_end.py
+++ b/colcon_core/event_handler/console_start_end.py
@@ -4,6 +4,8 @@
 import sys
 import time
 
+import colorama
+
 from colcon_core.event.job import JobEnded
 from colcon_core.event.job import JobStarted
 from colcon_core.event.test import TestFailure
@@ -25,6 +27,7 @@ class ConsoleStartEndEventHandler(EventHandlerExtensionPoint):
 
     def __init__(self):  # noqa: D107
         super().__init__()
+        colorama.init()
         satisfies_version(
             EventHandlerExtensionPoint.EXTENSION_POINT_VERSION, '^1.0')
         self._start_times = {}
@@ -35,7 +38,10 @@ class ConsoleStartEndEventHandler(EventHandlerExtensionPoint):
 
         if isinstance(data, JobStarted):
             self._start_times[data.identifier] = time.monotonic()
-            print(f'Starting >>> {data.identifier}', flush=True)
+            msg = ('Starting ' + colorama.Fore.GREEN +
+                   colorama.Style.BRIGHT + '>>>' + colorama.Fore.CYAN +
+                   f' {data.identifier}' + colorama.Style.RESET_ALL)
+            print(msg, flush=True)
 
         elif isinstance(data, TestFailure):
             job = event[1]
@@ -46,19 +52,30 @@ class ConsoleStartEndEventHandler(EventHandlerExtensionPoint):
                 time.monotonic() - self._start_times[data.identifier]
             duration_string = format_duration(duration)
             if not data.rc:
-                msg = f'Finished <<< {data.identifier} [{duration_string}]'
+                msg = (colorama.Style.BRIGHT + colorama.Fore.BLACK +
+                       'Finished ' + colorama.Fore.GREEN + '<<<'
+                       + colorama.Style.RESET_ALL + colorama.Fore.CYAN
+                       + f' {data.identifier}' + colorama.Fore.RESET
+                       + ' [' + colorama.Fore.YELLOW +
+                       f'{duration_string}' + colorama.Fore.RESET + ']')
                 job = event[1]
                 if job in self._with_test_failures:
                     msg += '\t[ with test failures ]'
                 writable = sys.stdout
 
             elif data.rc == SIGINT_RESULT:
-                msg = f'Aborted  <<< {data.identifier} [{duration_string}]'
+                msg = (colorama.Style.BRIGHT + colorama.Fore.RED +
+                       'Aborted  ' + colorama.Style.NORMAL + '<<<'
+                       + colorama.Fore.CYAN + f' {data.identifier}'
+                       + colorama.Fore.RESET)
                 writable = sys.stdout
-
             else:
-                msg = f'Failed   <<< {data.identifier} ' \
-                    f'[{duration_string}, exited with code {data.rc}]'
+                msg = (colorama.Style.BRIGHT + colorama.Fore.RED +
+                       'Failed   ' + colorama.Style.NORMAL + '<<<' +
+                       colorama.Fore.CYAN + f' {data.identifier}' +
+                       colorama.Fore.RESET + ' [' + colorama.Fore.RED +
+                       f'Exited with code {data.rc}' +
+                       colorama.Fore.RESET + ']')
                 writable = sys.stderr
 
             print(msg, file=writable, flush=True)
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,6 +27,7 @@ keywords = colcon
 [options]
 python_requires = >=3.6
 install_requires =
+  colorama
   coloredlogs; sys_platform == 'win32'
   distlib
   EmPy<4
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -1,6 +1,6 @@
 [colcon-core]
 No-Python2:
-Depends3: python3-distlib, python3-empy (<4), python3-packaging, python3-pytest, python3-setuptools, python3 (>= 3.8) | python3-importlib-metadata
+Depends3: python3-colorama, python3-distlib, python3-empy (<4), python3-packaging, python3-pytest, python3-setuptools, python3 (>= 3.8) | python3-importlib-metadata
 Recommends3: python3-pytest-cov
 Suggests3: python3-pytest-repeat, python3-pytest-rerunfailures
 Replaces3: colcon


v-lopez:
Thanks for this, I tested it and it is beautiful.

It may seem like a minor feature, but the first time I jumped from catkin tools to colcon, the lack of colors made the new tool even more daunting, as you haven't learnt yet which kind of messages are important and which can be ignored, so you have to read everything until you familiarize yourself with it. Colors bring a certain degree of familiarity, as well as highlighting important messages.

mikepurvis:
I definitely vote in favour of some strategy for coloring colcon, but I don't think the present PR addresses all of the concerns that have been raised in the past (eg #195):

- Will this gracefully degrade in the presence of a non-TERM, for example many CI setups, or running colcon piped directly to a file?
- How much is it going to complicate a future i18n effort?
- How badly is code readability going to be affected?
- What is the strategy for coordinating a rollout of this across colcon's sizable ecosystem of plugins?
- Is it okay to punt on "passing through" color codes from underlying tools (like make), or do we need a plan in place for that upfront? (eg, showing colors in the terminal view but stripping them from the saved logs)

Timple:
> the first time I jumped from catkin tools to colcon, the lack of colors made the new tool even more daunting

However minor such a feature sounds. I think the 'ROS2 migration fear' is a summation of such encounters. 
I think these kind of endeavors are really worth it!

v-lopez:
> Will this gracefully degrade in the presence of a non-TERM, for example many CI setups, or running colcon piped directly to a file? 
It looks like colorama is disabled when output is redirected. https://github.com/tartley/colorama#init-keyword-args

> How much is it going to complicate a future i18n effort?
I believe not much, since English sentences or words are not split by these tags or have different colors. 

> What is the strategy for coordinating a rollout of this across colcon's sizable ecosystem of plugins?
Like I said, I am new to colcon, and I'm not the author of this PR, but if you could enumerate them, I think these changes are simple enough so that all of us who are interested in this can push it easily.

>Is it okay to punt on "passing through" color codes from underlying tools (like make), or do we need a plan in place for that upfront? (eg, showing colors in the terminal view but stripping them from the saved logs)
I believe most of these tools already have systems in place to disable color output when redirecting. I know that I have to explicitly enable some of them to have colors in Jenkins output for instance.

Flova:
I am also highly in favor of this pull request! The missing color bothered me for quite some time now.

jaagut:
Hi! Thank you all for your great contributions!
I am just curious to know how to progress with this PR.
Are the open questions resolved? What should be discussed or changed? What stops this from being merged (except for the merge conflict)?

Flova:
Any updates on this PR?

Flova:
I think @v-lopez addressed the concerns quite well. Are there any other concerns that hold back colorization of colcon? @cottsay

jonselling:
Hi! I would really like to have colored output for colcon!

Are there any other concerns that need to be addressed?

MatthijsBurgh:
@timonegk @Timple @mikepurvis Lets put some effort to these color PRs.

Do we think the coloring should be in `colcon-core` or `colcon-output`?

mikepurvis:
I don't have enough of an architectural vision for colcon at this point— I think it has to be @cottsay makeing that call. Instinctively, it does feel like it needs to be in core, though, as colouring information is something that you can't really bolt onto the content after the fact; it has to be injected at the origin.

tokoro10g:
Any updates on this?

I agree with everyone here pointing out how the colorless output bothers developers.
I believe this PR will improve the DX of colcon pretty much and will encourage more people to use ROS2!

Timple:
Seems like this PR waiting long enough to have conflicts already :disappointed:

Flova:
Any updates on this?

Flova:
Fyi you can already easily use colcon with colors by executing the following commands. Keep in mind that no updates are applied that way so you need to execute them again if some interface changed and the fork updated.

```bash
pip install git+https://github.com/timonegk/colcon-core.git@colors --user
pip install git+https://github.com/timonegk/colcon-notification.git@colors --user
pip install git+https://github.com/timonegk/colcon-output.git@colors --user
```

MatthijsBurgh:
Friendly ping ;)

SammyRamone:
It has been more then 4 years now (including the original PR) and there is continous interest by multiple persons. Is there any way to get this PR finally merged? @cottsay

cottsay:
Please allow me to start by apologizing for neglecting this issue. There's really no excuse for not providing any input on these open PRs at all. I could tell you how thin resources are for maintaining colcon right now, and how there are significantly more impactful features and bugs that need attention, but I should have taken the time to communicate and iterate earlier.

---

Some of the reasons I've not felt that this is a priority feature are that there is very little console output generated by colcon itself, where much of the output comes from (already colorized) cmake output. Colorization also isn't universally usable by all users and in all contexts, and in some cases can even regress the experience, especially when colorization sequences appear in raw log files.

These are not excuses, but in the spirit of better communication, an explanation of some of my reluctance to take either of the open PRs as-is.

---

Here are some of my thoughts for moving this forward.

Functionality provided by `colcon-core` itself should be the bare minimum set of features for colcon to bootstrap itself. Even such trivial features as recursive directory traversal and package selection are implemented in separate extensions. So too should colorization of output, especially if it means picking up a new dependency. Less features means less bugs and simpler maintenance where it matters. However, this is complicated by the fact that there is at least some console output produced in colcon core itself, and you'll probably want to see colors there too.

Building from your desire for colorized output, I think we can build an efficient system to handle arbitrary output styling at little to no additional cost. An example of such augmentation would be section start/end markers such as those used by Travis CI, GitHub Actions, and Jenkins. Looking further out, I have created extensions for sending E-mail notifications for build failures that could take advantage of HTML for package URL hyperlinks, headers, and more.

Implementing this as an extension would also give users the opportunity to create "themes" and customize the output style to suit their preferences and environments without the need to modify existing packages.

I've created a branch with a proof-of-concept extension point for doing this: [cottsay/output-style](https://github.com/colcon/colcon-core/compare/cottsay/output-style)

Please take a look and tell me if you think this would suit your needs. I expect that we'd probably create a new extension, something like `colcon-style-colorama` that would map the high-level stylizers to specific ANSI codes in colorama, and we could consider adding that extension to the recommended family.

mikepurvis:
I definitely see the argument for adding hooks to do this and farming it out to a separate extension. However, I do wonder about whether the indirection associated with this is worth it. Colorama is well established and has already done the hard work of abstracting color codes across all of Python's supported platforms. At a certain point we risk making things more difficult and _less_ conventional by effectively having colcon-core roll its own in-house abstraction layer rather than leveraging what already exists.

At risk of attacking a straw man, I see the argument against native coloring as being not that far removed from arguing that colcon should abstract the terminal itself, or maybe even the concept of textual output. I know that's not really fair because those are things provided by the Python stdlib, but I think it's reasonable for Colorama's API to be natively part of the "platform" that colcon-core exposes to its ecosystem of plugins.

If nothing else, it's worth noting that this working PR adds 18 lines to `colcon-core` whereas the incomplete abstraction scheme being proposed adds ten times that, and would likely end up being much more once the plugin is also accounted for— that is an awfully high cost to pay for avoiding the coupling to a lightweight project that is well maintained and itself dependency-free.

tfoote:
Colorama in itself is moderately light weight and on it's own wouldn't be a blocker. And it's fair to consider whether the level of indirection is worth doing. However we can't turn around and then say that this implementation is very small compared to the cleaner abstraction and thus would be less maintenance when the current implementation is missing a lot of the functionality that is sketched into the abstraction. This change should be at least disableable to turn off colorized outputs. It's currently just hard coded. Furthermore, there's additionally no way to customize the colors. For example these hard coded colors are not accessible for red green color blind people. And if someone wants to do something more visually distinct or apply any other styling there's no room for extension. Without the abstraction extension we're now looking at adding this potential whole constellation of options to the core functionality. I know that some people strongly want this. If you'd like this please work to improve and flesh it out to provide the full functionality that @cottsay layed out which will unlock not just the single short term minimal use case but meet the larger vision for the tool, and not require deprecation cycles to get there.

mikepurvis:
Hmm, I can't think of any other tool of this class that provides color output and also has a mechanism for customizing the coloring— is there an example we can look at? Across `cmake`, `git`, `nix`, `brew`, `bazel`, and `gradle`, I think all of them do various color-y things without any ability to modify the scheme short of forking the source.

I suspect that for most, the pragmatic direction here would be:
- A pair of `--color`/`--no-color` flags, for which the default is on in the presence of a tty, and off otherwise. The "off" case is implemented by simply filtering the output right before printing it, likely doable through a log formatter.
- All plugins that submit output do so with the understanding that the color codes may be filtered.
- Color will never be used in a way that isn't redundant with the text itself, so it won't ever be _required_ to be able to distinguish (for example, red-ERROR/green-OK over red/green-PACKAGENAME). It's understood that users with special needs around specific color pairs will have the option to remap them from their terminal software if desired.

I do agree that it would be nice to have css-style separation between semantics and presentation, but in the absence of an established ecosystem implementing that for Python terminal output, I just don't know if it's worth building it from scratch inside colcon. Possibly a halfway measure could be simply defining a `Styles` enum in colcon-core with constants for common cases like `error`, `ok`, `info`, `emphasis`, `summary`, pointing directly to the colorama constants. Plugin authors can be encouraged to use those constants rather than hard-coding, and then there's at least single place to update the scheme down the road or make it more configurable/pluggable if that's later felt to be a requirement.

Flova:
I fully agree with it being a bit over the top to include advanced theming. Especially if the development is constrained and complexity should be low. Tbh. except things like `vim` that can be configured to be a whole IDE in the terminal, no CLI tools I know of provide that flexibility. 

I think it is important to be able to fall back to a non-colored mode easily. As discussed earlier in this thread `colorama` strips the coloring if the output is redirected (e.g. scripting or CI). An additional `--no-color` / `--color` flag might not hurt tho. and can be easily integrated in case you want colors in the CI or don't want them for aesthetic reasons.

I also agree with the other two points, especially that colors should be redundant.

Building custom stuff for standard things is also not ideal imo.. And especially if we have the choice between no feature, because we only implement a version that is harder to maintain and has additional things that are rarely used or a feature that is good enough for most users, I opt for the second one. Especially since anybody is still free to propose an extended version, that is integrated with proper deprecation or backwards compatibility. If this PR otherwise stays open for another 4 years, we can easily integrate a good enough (like most build tools use) solution now and deprecate it when the new one is ready. Even if that process takes a few years it's still faster, more user focused and not that much overhead. Also, I think the benefit of having proper formatting during normal use earlier is greater than pretty email formatting.

Additionally, each custom abstraction layer makes the coder harder to understand for new contributors etc.. Googling issues gets harder and the list goes on. This is not the first time I have noticed a tendency towards adding maybe a bit too many abstraction layers in ROS 2 which look good on paper, but in the end everybody uses the default one and it makes everything more painful when you change something or have a specific issue. Don't get me wrong, ROS 2 is awesome, but that is some criticism I have. Sometimes KISS is the way to go. 

tl;dr:

- We should keep things simple as most tools do
- Coloring should be easily disabled
- Custom stuff has its drawbacks and over engineering is a thing
- Too many can be abstractions are harmful


This being said, while I have some criticism, I thank @cottsay for his statement!

pulak-gautam:
Could anyone give an update on this? Has any extensible plugin been implemented which provides this functionality yet?

Timple:
I've been using this branch for months. Works like a charm.

However now it got out of date again. What is the way forward now? This PR has been open for years, let's get this in!

timonegk:
@Timple I rebased the branch on master, so at least for now it is up to date again.

Flova:
This seems to be stalled again ... 

While using the fork from @timonegk is nice, this is not a permanent solution. I would be happy if we could progress in some way.

nyxrobotics:
Thank you so much, it is a great feature.

spurnvoj:
Some progress has been made on the custom style: https://github.com/colcon/colcon-core/pull/687. However, I am unsure how to use it. @cottsay, can you help me with this? I also prefer the style that @timonegk proposed, which is heavily inspired by catkin_tools.

cottsay:
> This is beyond ridiculous. Anyone else down to start a new org for a community maintained fork of ‘colcon-core-colors’?

I added the output stylizer support in 0.19.0 in March, as @spurnvoj mentioned. You shouldn't need to fork colcon for this, but you do you.

I took a stab at what a catkin_tools style implementation would look like here: https://github.com/cottsay/colcon-ansi-colors-example

You'd select the stylizer with something like `colcon --output-style catkin_tools build`, but more likely you'd add that to your `colcon_defaults.yaml` so it's always used. It's possible for an extension to make the style enabled by default as well, using a higher priority.

If someone wants to refine my example and align it with what catkin_tools looked like, we can transfer it into the colcon org. You can also feel free to use colorama or any of the other popular colorizers if you feel like you need more Python dependencies in your life.

mikepurvis:
AFAICT `colcon_defaults.yaml` only lets you specify defaults for verb args, not global ones. I suspect this is a consequence of that functionality being itself supplied by a plugin, and so by the time the plugin is loaded and active, it's "too late" to interfere with global args:

https://github.com/colcon/colcon-defaults/blob/master/colcon_defaults/argument_parser/defaults.py#L64-L65

I'd love to be shown wrong on that, but in any case, the envvar is working great for me as well.

mcamurri:
> @cottsay, Thank you for the example! For those who don't know how to obtain it and try it, follow these instructions:
> 
> Install the colcon-ansi-colors-example extension using pip:
> 
> ```shell
> pip install git+https://github.com/cottsay/colcon-ansi-colors-example --break-system-packages
> ```
> 
> Once the extension is installed, you can use the new output style with the following command:
> 
> ```shell
> colcon  --output-style catkin_tools build
> ```
> 
> If you like the colorful output and want to use it by default for all colcon commands, add the following line to your shell's configuration file (e.g., `~/.bashrc` for Bash or `~/.zshrc` for Zsh):
> 
> ```shell
> export COLCON_DEFAULT_OUTPUT_STYLE=catkin_tools
> ```
> 
> After adding the line, be sure to open a new terminal or run `source ~/.bashrc` (or the appropriate file) for the change to take effect.
> 
> PS: I haven't managed to enable it using `colcon_defaults.yaml`.
> 
> Here is an example of the output: Selection_278

Thank you @spurnvoj for the instructions and @timonegk and everyone else for your contributions. 

Is there a way to get also the summary and progress being colorized? I'm on Jazzy and I saw [this](https://github.com/colcon/colcon-output/pull/43) and [this](https://github.com/colcon/colcon-output/pull/58), but I don't quite understand how to integrate them in my system.

The instructions above are the only ones I was able to understand and follow. I still find a pity that so much effort has to be made to get back a feature we already have 10 years ago by default.

Flova:
The following pip commands should set you up. On jazzy you might need to add an additional flag as writing in the system environment is protected in newer pip versions.

```
pip install git+https://github.com/timonegk/colcon-core.git@colors --user
pip install git+https://github.com/timonegk/colcon-notification.git@colors --user
pip install git+https://github.com/timonegk/colcon-output.git@colors --user
```

mcamurri:
Sorry, with all due respect, why isn't this PR not yet considered for merging? 

Colored output has been a basic feature of every terminal program for decades. The feature has been tested and works. Why isn't it yet integrated upstream?

The commands above seem not to be working anymore, most likely because colcon has changed since then:

```
ERROR:colcon.colcon_core.plugin_system:Exception instantiating extension 'colcon_core.executor.parallel': Extension point is too old (1.0), the extension requires '1.1'
  File "/home/mcamurri/.local/lib/python3.12/site-packages/colcon_core/plugin_system.py", line 60, in _instantiate_extension
    extension_instance = extension_class()
                         ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/colcon_parallel_executor/executor/parallel.py", line 47, in __init__
    satisfies_version(
  File "/home/mcamurri/.local/lib/python3.12/site-packages/colcon_core/plugin_system.py", line 175, in satisfies_version
    raise RuntimeError(
RuntimeError: Extension point is too old (1.0), the extension requires '1.1'

ERROR:colcon.colcon_core.extension_point:Exception loading extension 'colcon_core.event_handler.parallel_status': No module named 'colcon_core.output_style'
  File "/home/mcamurri/.local/lib/python3.12/site-packages/colcon_core/extension_point.py", line 169, in load_extension_points
    extension_type = load_extension_point(name, value, group)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mcamurri/.local/lib/python3.12/site-packages/colcon_core/extension_point.py", line 210, in load_extension_point
    return EntryPoint(name, value, group).load()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/importlib/metadata/__init__.py", line 205, in load
    module = import_module(match.group('module'))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 995, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/usr/lib/python3/dist-packages/colcon_parallel_executor/event_handler/parallel_status.py", line 7, in <module>
    from colcon_core.output_style import Style
ModuleNotFoundError: No module named 'colcon_core.output_style'
```

Still on Jazzy + Noble.

cottsay:
@mcamurri, it appears that you installed newer colcon extensions but the version of colcon-core you have installed is very old. It's likely because you're mixing packages provided by the debs (like `colcon_parallel_executor` is) with packages provided by pip (like `colcon-core` itself).

You should choose one or the other, and keep in mind that pip packages don't get updated when you run `apt upgrade`.

Timple:
OR (very weird suggestion), we merge this :smiley:

PR #496, opened 2022-03-23
Capture stdout and stderr

esteve:
This PR adds the output of running a process to the `stdout` and `stderr` fields of the returned [subprocess.CompletedProcess](https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess).

The rationale for this change is that because when running some build tools (e.g. Rust's `cargo`), they don't save their output to a file, so it's not possible to expose their results to `colcon` (e.g. saving test results for `colcon test-result`)

### DIFF
--- a/colcon_core/subprocess.py
+++ b/colcon_core/subprocess.py
@@ -72,12 +72,14 @@ async def run(
     assert callable(stderr_callback) or stderr_callback is None
 
     stdout_capture = []
+
     def _stdout_callback(line):
         if stdout_callback:
             stdout_callback(line)
         stdout_capture.append(line)
 
     stderr_capture = []
+
     def _stderr_callback(line):
         if stderr_callback:
             stderr_callback(line)
--- a/colcon_core/subprocess.py
+++ b/colcon_core/subprocess.py
@@ -48,6 +48,7 @@ async def run(
     stderr_callback: Callable[[bytes], None],
     *,
     use_pty: Optional[bool] = None,
+    capture_output: Optional[bool] = None,
     **other_popen_kwargs: Mapping[str, Any]
 ) -> subprocess.CompletedProcess:
     """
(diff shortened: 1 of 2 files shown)


mikepurvis:
I definitely thought this already happened. Are the logs in the `log` folder coming directly from the upstream tools vs being piped through colcon-core?

esteve:
@mikepurvis I believe the flow is that all output is redirected to `colcon`'s internal event queue which is then passed to the executor, which in turn dumps it into the `log` directory. However, nowhere in between you can access the process output unless you rewrite the `run` function from `colcon_core.subprocess` (see for example `check_call_return_stdout` in https://github.com/colcon/colcon-cargo/pull/4 as a workaround).

cottsay (changes requested):
I really don't like buffering all of the console output as a default behavior. This needs to be made conditional somehow, and the default behavior should not change.

Some ideas:
1. `subprocess.run` uses a `capture_output` argument to control this.
2. we could introduce special marker callbacks to pass to `std{out,err}_callback`, something like `colcon_core.subprocess.CAPTURE` or something. This has the added advantage of being inherently mutually exclusive with an explicit callback.
3. introduce a wrapper function to `run` that passes the wrapper callbacks to `run` and adds the captured output on the way out.

Don't get me wrong - I'm totally on board with supporting output capture somehow. I can already think of a handful of spots that will take advantage of it.

esteve:
@cottsay thanks for the feedback. Of the three options, I think the first one seems the simplest and most straightfoward, and IMHO the cleanest from a user's perspective, I'm going to implement that shortly and push it.

esteve:
@cottsay I've just pushed the changes implementing option 1 in 1218230e891e5d510ee58cf1ad50d97c58e5307d

cottsay (approved):
Thanks for iterating. This looks like a great solution.

esteve:
@cottsay thanks for the feedback and merging this PR :slightly_smiling_face:

PR #497, opened 2022-03-26
fix deprecated setuptools.config.read_configuration

ijnek:
Fix deprecations introduced in [setuptools V61.0.0](https://setuptools.pypa.io/en/latest/history.html#v61-0-0)

```
/home/user/colcon-venv/lib/python3.10/site-packages/setuptools/config/__init__.py:28: SetuptoolsDeprecationWarning: As setuptools moves its configuration towards `pyproject.toml`,
`setuptools.config.read_configuration` became deprecated.

For the time being, you can use the `setuptools.config.setupcfg` module
to access a backward compatible API, but this module is provisional
and might be removed in the future.
```

### DIFF
--- a/colcon_core/package_identification/python.py
+++ b/colcon_core/package_identification/python.py
@@ -86,7 +86,10 @@ def get_configuration(setup_cfg):
     """
     try:
         # import locally to allow other functions in this module to be usable
-        from setuptools.config.setupcfg import read_configuration
+        try:
+            from setuptools.config.setupcfg import read_configuration
+        except ImportError:
+            from setuptools.config import read_configuration
     except ImportError as e:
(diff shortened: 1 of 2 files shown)


ijnek:
@cottsay Thanks for the changes.

nuclearsandwich (approved):
From the setuptools changelog:

> Users that still need to parse and process configuration from setup.cfg can import a direct replacement from setuptools.config.setupcfg, however this module is transitional and might be removed in the future (the setup.cfg configuration format itself is likely to be deprecated in the future).

It seems like this will last us at least as long as the rest of the logic does. Thanks @ijnek for opening the PR and @cottsay for a review pass and improvements.

cottsay (approved):

PR #504, opened 2022-04-20
Use deb_system scheme when available to avoid 'local' in path

### LINKED ISSUE #503: Pip version of colcon unable to build ROS2 on Jammy
There is an issue with the `pip3` installed version of colcon which causes it to be unable to build ROS2 `rolling` on ubuntu `jammy`.

I have narrowed the issue to failing on `ament_cmake_core` with a particular set of `colcon` versions.

An example with a dockerfile that fails

https://gist.github.com/mjcarroll/a63c422fa8578a0999b5d8b20be76c96#file-dockerfile-bad

An example of a dockerfile pinned to specific `colcon` versions that works: 

https://gist.github.com/mjcarroll/a63c422fa8578a0999b5d8b20be76c96#file-dockerfile-good

To test:

```
docker build -t ros-rolling:jammy .
docker run -it ros-rolling:jammy /bin/bash
cd ~/ros2_rolling
colcon build --packages-up-to ament_cmake_core
```

Error output:
```
--- stderr: ament_cmake_core                         
  File "/root/ros2_rolling/src/ament/ament_cmake/ament_cmake_core/cmake/package_templates/templates_2_cmake.py", line 21, in <module>
    from ament_package.templates import get_environment_hook_template_path
ModuleNotFoundError: No module named 'ament_package'
CMake Error at ament_cmake_package_templates-extras.cmake:41 (message):
  execute_process(/usr/bin/python3.10
  /root/ros2_rolling/src/ament/ament_cmake/ament_cmake_core/cmake/package_templates/templates_2_cmake.py
  /root/ros2_rolling/build/ament_cmake_core/ament_cmake_package_templates/templates.cmake)
  returned error code 1
Call Stack (most recent call first):
  CMakeLists.txt:19 (include)


---
Failed   <<< ament_cmake_core [0.22s, exited with code 1]
```

```
$ colcon version-check
colcon-argcomplete 0.3.3: up-to-date
colcon-bash 0.4.2: up-to-date
colcon-cd 0.1.1: up-to-date
colcon-cmake 0.2.26: up-to-date
colcon-core 0.8.1: up-to-date
colcon-defaults 0.2.6: up-to-date
colcon-devtools 0.2.3: up-to-date
colcon-library-path 0.2.1: up-to-date
colcon-metadata 0.2.5: up-to-date
colcon-notification 0.2.13: up-to-date
colcon-output 0.2.12: up-to-date
colcon-package-information 0.3.3: up-to-date
colcon-package-selection 0.2.10: up-to-date
colcon-parallel-executor 0.2.4: up-to-date
colcon-pkg-config 0.1.0: up-to-date
colcon-powershell 0.3.7: up-to-date
colcon-python-setup-py 0.2.7: up-to-date
colcon-recursive-crawl 0.2.1: up-to-date
colcon-ros 0.3.23: up-to-date
colcon-test-result 0.3.8: up-to-date
colcon-zsh 0.4.0: up-to-date
```

sloretz:
Fixes #503

This PR makes colcon use `deb_system` scheme when getting python install paths if available. This works around the default `python_local` scheme on Ubuntu Jammy which was added in a Debian package patch. It does this with a new API `get_python_install_path` containing the logic for selecting the scheme.

I also added `PythonScriptsPathEnvironment` to get the python script install path. This should be `bin` on Linux if the scheme stuff is correct, but given that CPython packagers may mess with the schemes this gets it explicitly should that assumption not hold in the future.


cottsay (dismissed):
I believe this is the right fix - thanks for the investigation.

Please fix this in `ament` as well: https://github.com/ament/ament_cmake/blob/2bf27ce5ad01af340e1f4efd44232c1067d0dbda/ament_cmake_python/ament_cmake_python-extras.cmake#L49

cottsay:
I verified this behavior on Jammy, Fedora, and RHEL 8, and the behavior is consistent after this patch.

Interestingly, I could only replicate the bug on Jammy, even though Fedora 35 also uses Python 3.10.

sloretz:
> Interestingly, I could only replicate the bug on Jammy, even though Fedora 35 also uses Python 3.10.

Looks like there's a Debian patch adding `posix_local`

https://salsa.debian.org/cpython-team/python3/-/blob/master/debian/patches/sysconfig-debian-schemes.diff#L19

sloretz:
@cottsay I'm having some trouble with the bootstrap process on Jammy in an apptainer container. Would you be willing to check the bootstrap commands to see if I'm doing something wrong?

```console
$ apptainer build --fakeroot  bad.sif repro.def
[...]
+ eval export PATH="/root/colcon-from-source/install/colcon-core/bin:$PATH"
+ export PATH=/root/colcon-from-source/install/colcon-core/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+ unset _colcon_ordered_commands
+ unset _colcon_prefix_sh_source_script
+ unset _colcon_prefix_sh_COLCON_CURRENT_PREFIX
+ colcon build
  File "/root/colcon-from-source/install/colcon-core/bin/colcon", line 33, in <module>
    sys.exit(load_entry_point('colcon-core==0.8.1', 'console_scripts', 'colcon')())
  File "/root/colcon-from-source/install/colcon-core/bin/colcon", line 22, in importlib_load_entry_point
    for entry_point in distribution(dist_name).entry_points
  File "/usr/lib/python3.10/importlib/metadata/__init__.py", line 957, in distribution
    return Distribution.from_name(distribution_name)
  File "/usr/lib/python3.10/importlib/metadata/__init__.py", line 548, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: No package metadata was found for colcon-core
FATAL:   While performing build: while running engine: exit status 1
```

<details><summary>repro.def</summary>

```
Bootstrap: docker
From: ubuntu:22.04

%post
  export DEBIAN_FRONTEND=noninteractive

  apt update
  apt install locales tzdata
  locale-gen en_US en_US.UTF-8
  update-locale LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8

  echo 'Etc/UTC' > /etc/timezone
  ln -sf /usr/share/zoneinfo/Etc/UTC /etc/localtime

  apt install -y --quiet --no-install-recommends \
    build-essential \
    cmake \
    git \
    python3-flake8 \
    python3-flake8-blind-except \
    python3-flake8-builtins \
    python3-flake8-class-newline \
    python3-flake8-comprehensions \
    python3-flake8-deprecated \
    python3-flake8-docstrings \
    python3-flake8-import-order \
    python3-flake8-quotes \
    python3-pip \
    python3-pytest \
    python3-pytest-cov \
    python3-pytest-repeat\
    python3-pytest-rerunfailures \
    python3-setuptools \
    wget \
    curl

  python3 -m pip install -U vcstool pip setuptools

  # Broken when installed from pip
  # python3 -m pip install -U colcon-common-extensions

  # Broken when bootstrapped from source
  mkdir ~/colcon-from-source && cd ~/colcon-from-source
  curl --output colcon.repos https://raw.githubusercontent.com/colcon/colcon.readthedocs.org/main/colcon.repos
  mkdir src
  vcs import src < colcon.repos

  # Not broken when bootstrapped with scheme fix
  cd src/colcon-core
  git checkout sloretz__sysconfig__scheme
  cd -

  curl --output requirements.txt https://raw.githubusercontent.com/colcon/colcon.readthedocs.org/main/requirements.txt
  pip install -r requirements.txt
  ./src/colcon-core/bin/colcon build --paths src/*
  . install/local_setup.sh
  colcon build

  mkdir -p ~/ros2_rolling/src
  cd ~/ros2_rolling
  wget https://raw.githubusercontent.com/ros2/ros2/master/ros2.repos
  vcs import src < ros2.repos

%environment
  export LANG=en_US.UTF-8
  export ROS_PYTHON_VERSION=3
```
</details>

cottsay:
> Looks like there's a Debian patch adding `posix_local`

I always figured this problem was `setuptools`' fault, but you're right, it looks like that patch is what's causing all this nonsense.

cottsay:
> I'm having some trouble with the bootstrap process on Jammy

This PR is missing an update to this use of `sysconfig.get_path`:
https://github.com/colcon/colcon-core/blob/ac294dfd480f91495df991989493f777a62e08fb/colcon_core/task/python/build.py#L298-L300

That seemed to resolve the problem for me.

sloretz:
> This PR is missing an update to this use of sysconfig.get_path:

That brings it a little closer, but there's still al local path `install/colcon-core/local/bin/colcon` here preventing me from bootstrapping because `colcon` isn't on `PATH` after sourcing the setup script.

clalancette:
The latest change here (along with https://github.com/ament/ament_cmake/pull/386) fixed the original problem for me.  So I would say we are making progress.

cottsay:
> > This PR is missing an update to this use of sysconfig.get_path:
> 
> That brings it a little closer, but there's still al local path `install/colcon-core/local/bin/colcon` here preventing me from bootstrapping because `colcon` isn't on `PATH` after sourcing the setup script.

I just tried from scratch, and I can no longer reproduce your error after you pushed the latest fix.

Here's the spot that needs to be updated to resolve the test failure: https://github.com/colcon/colcon-core/blob/d2882384b42b065f511ce4e98759e9a5597ffd1d/test/test_environment_pythonpath.py#L25-L30

cottsay:
Your new `PythonScriptsPathEnvironment` shouldn't be necessary. sysconfig.get_path('scripts') shouldn't be returning anything different from `<prefix>/bin`, so the normal `PathEnvironment` should be finding the scripts and appending to `PATH`.

I can't reproduce your `PATH` problem. Please give me repro steps.

sloretz:
> I can't reproduce your PATH problem. Please give me repro steps.

Here they are. Maybe it's a newer version of setuptools using `sysconfig.get_path` internally? If so, maybe embracing the `local` is the way to go.

<details><summary>repro_binprob.def</summary>

```
Bootstrap: docker
From: ubuntu:22.04

%post
  export DEBIAN_FRONTEND=noninteractive

  apt update
  apt install locales tzdata
  locale-gen en_US en_US.UTF-8
  update-locale LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8

  echo 'Etc/UTC' > /etc/timezone
  ln -sf /usr/share/zoneinfo/Etc/UTC /etc/localtime

  apt install -y --quiet --no-install-recommends \
    build-essential \
    cmake \
    git \
    python3-flake8 \
    python3-flake8-blind-except \
    python3-flake8-builtins \
    python3-flake8-class-newline \
    python3-flake8-comprehensions \
    python3-flake8-deprecated \
    python3-flake8-docstrings \
    python3-flake8-import-order \
    python3-flake8-quotes \
    python3-pip \
    python3-pytest \
    python3-pytest-cov \
    python3-pytest-repeat\
    python3-pytest-rerunfailures \
    python3-setuptools \
    wget \
    curl

  python3 -m pip install -U vcstool pip setuptools

  # Broken when installed from pip
  # python3 -m pip install -U colcon-common-extensions

  # Broken when bootstrapped from source
  mkdir /colcon-from-source && cd /colcon-from-source
  curl --output colcon.repos https://raw.githubusercontent.com/colcon/colcon.readthedocs.org/main/colcon.repos
  mkdir src
  vcs import src < colcon.repos

  # Not broken when bootstrapped with scheme fix
  cd src/colcon-core
  git checkout sloretz__sysconfig__scheme
  cd -

  curl --output requirements.txt https://raw.githubusercontent.com/colcon/colcon.readthedocs.org/main/requirements.txt
  pip install -r requirements.txt
  ./src/colcon-core/bin/colcon build --paths src/*
  . install/local_setup.sh
  echo $PYTHONPATH
  colcon build

  # Problem!
  # colcon is installed to `<prefix>/local/bin/colcon`

%environment
  export LANG=en_US.UTF-8
  export ROS_PYTHON_VERSION=3
```
</details>

Problem is `colcon` is installed to `local/bin/colcon` instead of `bin/colcon`

```console
$ apptainer build --fakeroot binprob.sif repro_binprob.def
[...]
+ colcon build
/.post.script: 54: colcon: not found
```

cottsay:
> Here they are.

Thank you. I see the issue now.

Is there a corresponding patch to the `python3-setuptools` deb that gets overridden when we use `setuptools` from pip? It would be nice to understand why the behavior keeps flip-flopping like this.

> maybe embracing the `local` is the way to go.

Yeah, that might be a good plan, in which case your `PythonScriptsPathEnvironment` will be needed. We'd also need to add an `ament` hook for non-colcon builds.

It's going to be strange having the executables split between `bin` and `local/bin` on Jammy like this. There will probably be fallout from code making assumptions.

sloretz:
> Is there a corresponding patch to the python3-setuptools deb that gets overridden when we use setuptools from pip? It would be nice to understand why the behavior keeps flip-flopping like this.

This patch in Jammy setuptools seems relevant (I couldn't find a web link)

combined with this code
https://github.com/colcon/colcon-core/blob/37cf8aaee99b9a65c8dc5579abc7d43c631a1f97/colcon_core/task/python/build.py#L302-L304

<details><summmary>install-layout.diff</summary>

```patch
--- a/setuptools/command/easy_install.py
+++ b/setuptools/command/easy_install.py
@@ -138,6 +138,8 @@ class easy_install(Command):
         ('local-snapshots-ok', 'l',
          "allow building eggs from local checkouts"),
         ('version', None, "print version information and exit"),
+        ('install-layout=', None, "installation layout to choose (known values: deb)"),
+        ('force-installation-into-system-dir', '0', "force installation into /usr"),
         ('no-find-links', None,
          "Don't load find-links defined in packages being installed"),
         ('user', None, "install in user site-package '%s'" % site.USER_SITE)
@@ -145,7 +147,7 @@ class easy_install(Command):
     boolean_options = [
         'zip-ok', 'multi-version', 'exclude-scripts', 'upgrade', 'always-copy',
         'editable',
-        'no-deps', 'local-snapshots-ok', 'version',
+        'no-deps', 'local-snapshots-ok', 'version', 'force-installation-into-system-dir'
         'user'
     ]
 
@@ -188,6 +190,10 @@ class easy_install(Command):
         self.pth_file = self.always_copy_from = None
         self.site_dirs = None
         self.installed_projects = {}
+        # enable custom installation, known values: deb
+        self.install_layout = None
+        self.force_installation_into_system_dir = None
+
         # Always read easy_install options, even if we are subclassed, or have
         # an independent instance created.  This ensures that defaults will
         # always come from the standard configuration file(s)' "easy_install"
@@ -259,6 +265,11 @@ class easy_install(Command):
         self.expand_basedirs()
         self.expand_dirs()
 
+        if self.install_layout:
+            if not self.install_layout.lower() in ['deb']:
+                raise DistutilsOptionError("unknown value for --install-layout")
+            self.install_layout = self.install_layout.lower()
+
         self._expand(
             'install_dir', 'script_dir', 'build_directory',
             'site_dirs',
@@ -285,6 +296,15 @@ class easy_install(Command):
         if self.user and self.install_purelib:
             self.install_dir = self.install_purelib
             self.script_dir = self.install_scripts
+
+        if self.prefix == '/usr' and not self.force_installation_into_system_dir:
+            raise DistutilsOptionError("""installation into /usr
+
+Trying to install into the system managed parts of the file system. Please
+consider to install to another location, or use the option
+--force-installation-into-system-dir to overwrite this warning.
+""")
+
         # default --record from the install command
         self.set_undefined_options('install', ('record', 'record'))
         # Should this be moved to the if statement below? It's not used
@@ -1322,11 +1342,28 @@ class easy_install(Command):
                 self.debug_print("os.makedirs('%s', 0o700)" % path)
                 os.makedirs(path, 0o700)
 
+    if sys.version[:3] in ('2.3', '2.4', '2.5') or 'real_prefix' in sys.__dict__:
+        sitedir_name = 'site-packages'
+    else:
+        sitedir_name = 'dist-packages'
+
     INSTALL_SCHEMES = dict(
         posix=dict(
             install_dir='$base/lib/python$py_version_short/site-packages',
             script_dir='$base/bin',
         ),
+        unix_local = dict(
+            install_dir = '$base/local/lib/python$py_version_short/%s' % sitedir_name,
+            script_dir  = '$base/local/bin',
+        ),
+        posix_local = dict(
+            install_dir = '$base/local/lib/python$py_version_short/%s' % sitedir_name,
+            script_dir  = '$base/local/bin',
+        ),
+        deb_system = dict(
+            install_dir = '$base/lib/python3/%s' % sitedir_name,
+            script_dir  = '$base/bin',
+        ),
     )
 
     DEFAULT_SCHEME = dict(
@@ -1337,11 +1374,18 @@ class easy_install(Command):
     def _expand(self, *attrs):
         config_vars = self.get_finalized_command('install').config_vars
 
-        if self.prefix:
+        if self.prefix or self.install_layout:
+            if self.install_layout and self.install_layout in ['deb']:
+                    scheme_name = "deb_system"
+                    self.prefix = '/usr'
+            elif self.prefix or 'real_prefix' in sys.__dict__:
+                scheme_name = os.name
+            else:
+                scheme_name = "posix_local"
             # Set default install_dir/scripts from --prefix
             config_vars = config_vars.copy()
             config_vars['base'] = self.prefix
-            scheme = self.INSTALL_SCHEMES.get(os.name, self.DEFAULT_SCHEME)
+            scheme = self.INSTALL_SCHEMES.get(scheme_name,self.DEFAULT_SCHEME)
             for attr, val in scheme.items():
                 if getattr(self, attr, None) is None:
                     setattr(self, attr, val)
@@ -1385,9 +1429,15 @@ def get_site_dirs():
             sitedirs.extend([
                 os.path.join(
                     prefix,
+                    "local/lib",
+                    "python" + sys.version[:3],
+                    "dist-packages",
+                ),
+                os.path.join(
+                    prefix,
                     "lib",
                     "python{}.{}".format(*sys.version_info),
-                    "site-packages",
+                    "dist-packages",
                 ),
                 os.path.join(prefix, "lib", "site-python"),
             ])
--- a/setuptools/command/install_egg_info.py
+++ b/setuptools/command/install_egg_info.py
@@ -1,5 +1,5 @@
 from distutils import log, dir_util
-import os
+import os, sys
 
 from setuptools import Command
 from setuptools import namespaces
@@ -18,14 +18,31 @@ class install_egg_info(namespaces.Instal
 
     def initialize_options(self):
         self.install_dir = None
+        self.install_layout = None
+        self.prefix_option = None
 
     def finalize_options(self):
         self.set_undefined_options('install_lib',
                                    ('install_dir', 'install_dir'))
+        self.set_undefined_options('install',('install_layout','install_layout'))
+        if sys.hexversion > 0x2060000:
+            self.set_undefined_options('install',('prefix_option','prefix_option'))
         ei_cmd = self.get_finalized_command("egg_info")
         basename = pkg_resources.Distribution(
             None, None, ei_cmd.egg_name, ei_cmd.egg_version
         ).egg_name() + '.egg-info'
+
+        if self.install_layout:
+            if not self.install_layout.lower() in ['deb']:
+                raise DistutilsOptionError("unknown value for --install-layout")
+            self.install_layout = self.install_layout.lower()
+            basename = basename.replace('-py%s' % pkg_resources.PY_MAJOR, '')
+        elif self.prefix_option or 'real_prefix' in sys.__dict__:
+            # don't modify for virtualenv
+            pass
+        else:
+            basename = basename.replace('-py%s' % pkg_resources.PY_MAJOR, '')
+
         self.source = ei_cmd.egg_info
         self.target = os.path.join(self.install_dir, basename)
         self.outputs = []
```
</details>

I'm going to grab lunch, but now looking at that setuptools patch I'm thinking another option is to use the scheme `deb_system` if it exists (instead of `posix_prefix`). Or maybe another option is to not change the `--install-layout` in the Python build task.

sloretz:
I updated the PR with a solution using `deb_system` when available, and I've tested it locally through the bootstrap process.

colcon_core/python_install_path.py
```diff
@@ -0,0 +1,33 @@
...
+        preferred_scheme = sysconfig.get_preferred_scheme('prefix')
+    else:
+        # Python < 3.10
+        preferred_scheme = sysconfig._get_default_scheme()
+
+    if 'deb_system' in sysconfig.get_scheme_names():
+        # Ubuntu Jammy has a custom scheme
+        preferred_scheme = 'deb_system'
+
+    return Path(sysconfig.get_path(name, preferred_scheme, vars_))
```
cottsay, line 33: We can do this without relying on non-public API:
```suggestion
    kwargs = {}
    if vars_:
      kwargs['vars'] = vars_
    if 'deb_system' in sysconfig.get_scheme_names():
        kwargs['scheme'] = 'deb_system'
        
    return Path(sysconfig.get_path(name, **kwargs))
```

colcon_core/environment/path.py
```diff
@@ -45,3 +46,41 @@ def _create_environment_hooks(
...
+    def __init__(self):  # noqa: D107
+        super().__init__()
+        satisfies_version(
+            EnvironmentExtensionPoint.EXTENSION_POINT_VERSION, '^1.0')
+
+    def create_environment_hooks(self, prefix_path, pkg_name):  # noqa: D102
+        hooks = self._create_environment_hooks(prefix_path, pkg_name, 'bin')
+        if sys.platform == 'win32':
+            hooks += self._create_environment_hooks(
+                prefix_path, pkg_name, 'Scripts', '-scripts')
```
cottsay, line 63: This is copy/paste cruft from the original `PathEnvironment` class, and it can actually be removed from there too because `PythonScriptsPathEnvironment` actually obsoletes it. So both of these extensions can be simplified a fair amount.

I verified that `sysconfig.get_path('scripts', vars={'base': '/foo'})` returns `/foo/Scripts` on Windows 11 with Python 3.9.

colcon_core/environment/path.py
```diff
@@ -45,3 +46,41 @@ def _create_environment_hooks(
...
+            names = os.listdir(str(bin_path))
+        except FileNotFoundError:
+            pass
+        else:
+            for name in names:
+                if not (bin_path / name).is_file():
+                    continue
+                rel_bin_path = bin_path.relative_to(prefix_path)
+                hooks += shell.create_environment_hook(
+                    'path' + suffix, prefix_path, pkg_name, 'PATH',
```
cottsay, line 82: I believe that the first argument to `create_environment_hook` is an ID that is expected to be unique, so I believe it should be different from what's passed in the `PathEnvironment` extension.

setup.cfg
```diff
@@ -80,6 +80,7 @@ python_classes = !TestFailure
 colcon_core.argument_parser =
 colcon_core.environment =
     path = colcon_core.environment.path:PathEnvironment
+    pythonscriptspath = colcon_core.environment.path:PythonScriptsPathEnvironment
     pythonpath = colcon_core.environment.pythonpath:PythonPathEnvironment
```
cottsay, line 84:
```suggestion
    pythonpath = colcon_core.environment.pythonpath:PythonPathEnvironment
    pythonscriptspath = colcon_core.environment.path:PythonScriptsPathEnvironment
```

test/spell_check.words
```diff
@@ -119,6 +120,7 @@ todo
 traceback
 tryfirst
 tuples
+ubuntu
```
cottsay, line 123:
```suggestion
```

cottsay (changes requested):

colcon_core/environment/path.py
sloretz: I used `pythonscriptspath` as the id in one of the recent commits

colcon_core/environment/path.py
sloretz: Simplified logic in the recent commits. Let me know what you think!

cottsay: Loads better. Thank you!

colcon_core/python_install_path.py
```diff
@@ -0,0 +1,25 @@
...
+    See sysconfig.get_path for more info about the arguments.
+
+    :param name: Name of the path type
+    :param vars_: A dictionary of variables updating the values of
+        sysconfig.get_config_vars()
+    :rtype: Pathlib.Path
+    """
+    kwargs = {}
+    if vars_:
+        kwargs['vars'] = vars_
```
cottsay, line 21: Do you mind if we drop this for the sake of test coverage? The case is handled in CPython:
https://github.com/python/cpython/blob/1748816e80b23744667e239b49b477c0e283d201/Lib/sysconfig.py#L214-L218
```suggestion
    kwargs = {'vars': vars_}
```

sloretz: In 6329f54 I took a avoided the conditional with a tuple to avoid assuming `sysconfig` will accept `vars` being `None`, since the documentation says `If vars is provided, it must be a dictionary`.

cottsay:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16597/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11179/)
* Linux-RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/220/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16985/)

cottsay (approved):
Thanks for iterating with me on this - I'm really happy with how this change turned out. The separated `PythonScriptsPathEnvironment`, though probably not necessary after changing to use `deb_system` opportunistically, should help reduce the scope of issues like this in the future, and it removed the explicit workaround for Windows.

PR #658, opened 2024-07-31
Support empy4 and empy3

ahcorde:
Added some changes to support empy3 and empy4

@cottsay do you mind to update your branch with `master` ?


ahcorde:
@clalancette FYI

PR #713, opened 2025-10-21
Enhance tests for 'sh' and 'bat' shells

cottsay:
The existing tests exercise much of the code for implementing colcon's shell subsystem, but it doesn't validate the results. If the platform supports the shell, we should check that the code for appending and prepending paths to lists function as intended.

In each test module, the first block of changes is just a refactor so that we reference the hooks generated by the calls to `create_hook_*`. Previously, the test only validated that the hook could be generated, and didn't attempt to actually use it.

The second block adds a handful of assertions about the behavior of the hooks when executed.

### DIFF
--- a/test/test_shell_bat.py
+++ b/test/test_shell_bat.py
@@ -208,10 +208,10 @@ def _test_prefix_script(prefix_path):
         subdirectory_path = str(prefix_path / 'subdirectory')
 
         # validate appending/prepending without existing values
-        with patch.dict(os.environ, {
-            'APPEND_NAME': '',
-            'PREPEND_NAME': '',
-        }):
+        with patch.dict(os.environ) as env_patch:
+            env_patch.pop('APPEND_NAME', None)
+            env_patch.pop('PREPEND_NAME', None)
+
             coroutine = _run_prefix_script(prefix_script)
             env = run_until_complete(coroutine)
             assert env.get('APPEND_NAME') == subdirectory_path
--- a/test/test_shell_sh.py
+++ b/test/test_shell_sh.py
@@ -210,10 +210,10 @@ def _test_prefix_script(prefix_path):
         subdirectory_path = str(prefix_path / 'subdirectory')
 
         # validate appending/prepending without existing values
-        with patch.dict(os.environ, {
-            'APPEND_NAME': '',
-            'PREPEND_NAME': '',
-        }):
+        with patch.dict(os.environ) as env_patch:
+            env_patch.pop('APPEND_NAME', None)
+            env_patch.pop('PREPEND_NAME', None)
+
             coroutine = _run_prefix_script(prefix_script)
             env = run_until_complete(coroutine)
             assert env.get('APPEND_NAME') == subdirectory_path


test/test_shell_bat.py
```diff
@@ -39,33 +44,36 @@ def _test_extension(prefix_path):
     extension.create_prefix_script(prefix_path, False)
     assert (prefix_path / 'local_setup.bat').exists()
 
+    # create_hook_append_value
```
knmcguire: These are tests that were already included, but the only thing that changed was that now the hook_path is changed to 2 separate values to be used later.

test/test_shell_bat.py
```diff
@@ -39,33 +44,36 @@ def _test_extension(prefix_path):
...
+        'APPEND_NAME', 'subdirectory')
+    assert append_hook_path.exists()
+    assert append_hook_path.name == 'append_env_hook_name.bat'
+    content = append_hook_path.read_text()
+    assert 'APPEND_NAME' in content
+
+    # create_hook_prepend_value
+    prepend_hook_path = extension.create_hook_prepend_value(
+        'prepend_env_hook_name', prefix_path, 'pkg_name',
+        'PREPEND_NAME', 'subdirectory')
```
knmcguire: To rename "NAME" and "env_hook_name.bat" would also be a better cleanup indeed.

test/test_shell_bat.py
```diff
@@ -39,33 +44,36 @@ def _test_extension(prefix_path):
...
+    assert prepend_hook_path.exists()
+    assert prepend_hook_path.name == 'prepend_env_hook_name.bat'
+    content = prepend_hook_path.read_text()
+    assert 'PREPEND_NAME' in content
+
     # create_package_script
     extension.create_package_script(
         prefix_path, 'pkg_name', [
-            ('hookA.bat', '/some/path/hookA.bat'),
+            (append_hook_path.relative_to(prefix_path), ()),
```
knmcguire: Good efficient use of the earlier made temporary hookfiles to use for this test, which also tests if no 'non bat files' are added to the path as well.

test/test_shell_bat.py
```diff
@@ -93,3 +101,145 @@ def _test_extension(prefix_path):
...
+        with patch.dict(os.environ) as env_patch:
+            env_patch.pop('APPEND_NAME', None)
+            env_patch.pop('PREPEND_NAME', None)
+            coroutine = extension.generate_command_environment(
+                'task_name', prefix_path, {'pkg_name': str(prefix_path)})
+            env = run_until_complete(coroutine)
+            assert env.get('APPEND_NAME') == subdirectory_path
+            assert env.get('PREPEND_NAME') == subdirectory_path
+
+        # validate appending/prepending with existing values
```
knmcguire: So here I perhaps don't really understand what is going on. As I see

```
        with patch.dict(os.environ, {
            'APPEND_NAME': 'control',
            'PREPEND_NAME': 'control',
        }):
```
Replaces the paths with just 'control', but the asserts:

```
            assert env.get('APPEND_NAME') == os.pathsep.join((
                'control',
                subdirectory_path,
            ))
            assert env.get('PREPEND_NAME') == os.pathsep.join((
                subdirectory_path,
                'control',
            ))
```
checks if they contain the subdirectory_path + 'control'. Wouldn't this always assert?

Update: Ah wait now I see it. the append and prepend hook scripts are supposed to add the subdirectory path to the append and prepend paths.

test/test_shell_bat.py
```diff
@@ -93,3 +101,145 @@ def _test_extension(prefix_path):
...
+            assert env.get('APPEND_NAME') == os.pathsep.join((
+                'control',
+                subdirectory_path,
+            ))
+            assert env.get('PREPEND_NAME') == os.pathsep.join((
+                subdirectory_path,
+                'control',
+            ))
+
+        # validate appending/prepending unique values
```
knmcguire: And here the tests test that, if the subdirectory_path has been added twice, that the generate_command_environment won't add that a second time (so that it stays unique)

test/test_shell_bat.py
```diff
@@ -93,3 +101,145 @@ def _test_extension(prefix_path):
...
+            (prepend_hook_path.relative_to(prefix_path), ())])
+    assert (prefix_path / 'share' / 'pkg_name' / 'package.dsv').exists()
+    content = (prefix_path / 'share' / 'pkg_name' / 'package.dsv').read_text()
+    assert append_hook_path.name in content
+    assert prepend_hook_path.name in content
+
+    # create_prefix_script
+    extension.create_prefix_script(prefix_path, True)
+    prefix_script = prefix_path / 'local_setup.bat'
+    prefix_script.exists()
```
knmcguire: Wait... This might be mistake. Shouldn't `prefix_script.exists()` be an assert as well?

test/test_shell_bat.py
```diff
@@ -93,3 +101,145 @@ def _test_extension(prefix_path):
...
+    assert (prefix_path / '_local_setup_util_bat.py').exists()
+
+    if sys.platform == 'win32':
+        subdirectory_path = str(prefix_path / 'subdirectory')
+
+        # validate appending/prepending without existing values
```
knmcguire: Same test as the one int _test_extensions, but now with local_setup.bat. As are the 'existing values' and 'unique values' tests.

test/test_shell_sh.py
```diff
@@ -94,3 +98,158 @@ def _test_extension(prefix_path):
...
+            (prepend_hook_path.relative_to(prefix_path), ())])
+    assert (prefix_path / 'share' / 'pkg_name' / 'package.dsv').exists()
+    content = (prefix_path / 'share' / 'pkg_name' / 'package.dsv').read_text()
+    assert append_hook_path.name in content
+    assert prepend_hook_path.name in content
+
+    # create_prefix_script
+    extension.create_prefix_script(prefix_path, True)
+    prefix_script = prefix_path / 'local_setup.sh'
+    prefix_script.exists()
```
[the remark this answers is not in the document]
knmcguire:
```suggestion
    assert prefix_script.exists()
```

test/test_shell_bat.py
knmcguire:
```suggestion
    assert prefix_script.exists()
```

knmcguire:
Also another question, perhaps more of a design question than about this PR... how about the local_setup.ps1 scripts for powershell? those are generated but not tested I see.

cottsay:
> Also another question, perhaps more of a design question than about this PR... how about the local_setup.ps1 scripts for powershell? those are generated but not tested I see.

Shells other than `sh` and `bat` are part of separate extensions, such as this one: https://github.com/colcon/colcon-powershell

They should have their own tests. It's very likely that the additional checks I'm adding here are appropriate for those other shells as well, but I'm not sure I have the bandwidth to port them over any time soon. Help Wanted :grin:

test/test_shell_bat.py
cottsay: 👍 Same assertions, different process.

PR #714, opened 2025-10-21
Conform to linters in prefix_util.py.em

cottsay:
Because this is a template, it isn't automatically checked by our linters. This change cleans up a handful of violations and should not result in any change in behavior.

Namely:
- line length
- spelling
- unnecessary scope qualifiers

Requires tests that are added in #713


knmcguire:
Hmm.. I did notice that the tests for this PR don't fail, while those for #713  do fail due to finding those leading and trailing separators. This branch should have been based on cottsay/shell-tests... so why is it not detecting those now 🤔 or are the branches now different?

PR #726, opened 2026-02-24
Allow declaration of "out-of-band" dependencies

cottsay:
The "out-of-band" declaration on a dependency is intended to indicate that the dependency is a requirement but not a prerequisite for processing. As such, operations like "--packages-up-to" and graph computation will still show the relationship, but the topological ordering operation will not require the dependency as a prerequisite.

### DIFF
--- a/colcon_core/topological_order.py
+++ b/colcon_core/topological_order.py
@@ -41,14 +41,7 @@ def topological_order_decorators(decorators):
     for decorator in decorators:
         queued[decorator] = {
             d.name for d in decorator.recursive_dependencies
-        }
-
-    # map the set of out-of-band dependencies for each decorator
-    out_of_band = {}
-    for decorator in decorators:
-        out_of_band[decorator] = {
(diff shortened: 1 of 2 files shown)


colcon_core/topological_order.py
```diff
@@ -43,15 +43,37 @@ def topological_order_decorators(decorators):
...
+            ready = [
+                decorator for decorator, r in queued.items()
+                if not r.difference(out_of_band[decorator])
+            ]
+            # prefer those with less out-of-band dependencies first
+            ready.sort(key=lambda d: (
+                len(out_of_band[d]), d.descriptor.name
+            ))
+            # take only one at a time
+            ready = ready[:1]
```
Blast545: Should this section be optional with a condition somehow? The out-of-band concept is a bit tricky to grasp, it would be nice to add an example to illustrate what we want. 

We discussed it in a meeting, but it's not clear to me from the PR's code completely how this will work.

PR #728, opened 2026-03-12
Add framework for enumerating installed Python files

cottsay:
In order for colcon to consume wheels produced by PEP 517 build backends, we'll need to explicitly uninstall existing Python packages from the install space. This primarily stems from the differences in metadata representation, where the previous setuptools-only Python build mechanisms wrote `.egg-link` and `.egg-info` metadata, and wheels will use `.dist-info` metadata.

There should be only one metadata representation in the install space for a given package. The legacy setuptools build logic already has a special case for dealing with a switch from `.egg-info` to `.egg-link` and vice-versa. As we toss `.dist-info` into the mix, the need arises for a more robust solution.

This new functionality is not currently used by colcon, but will be used as part of the transition to standards-based Python build support.

> [!NOTE]
> This looks like a lot of code. While it isn't a trivial change, all but one of the files are tests, and even that one file contains a significant amount of boilerplate for integrating with the `importlib.metadata` mechanisms for distribution discovery.
> It's well-tested and thoroughly implemented because it will be a critical component of colcon's standards-based Python package building infrastructure.

### DIFF
--- a/colcon_core/dependency_descriptor.py
+++ b/colcon_core/dependency_descriptor.py
@@ -19,11 +19,14 @@ class DependencyDescriptor(str):
     """
 
     @staticmethod
-    def __new__(cls, name, *, metadata=None):  # noqa: D102
+    def __new__(cls, name, *args, **kwargs):  # noqa: D102
         return str.__new__(cls, name)
 
-    def __init__(self, name, *, metadata=None):  # noqa: D107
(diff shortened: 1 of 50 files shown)


KmoM88 (approved):
LGTM, Thank you! I'll add a reference of this framework in the proposal issue #736

PR #729, opened 2026-03-31
Unify temporary directory allocation strategy in tests

cottsay:
This package was previously using three different temporary directory allocation strategies:
- Manual `TemporaryDirectory` management
- pytest `tmpdir` fixture
- pytest `tmp_path` fixture

This commit changes all temporary directory allocation in tests to use the `tmp_path` fixture, and includes a compatibility definition of that fixture so that the tests will continue to work properly on Enterprise Linux 8.

---

I'm not expecting a thorough review of this change. Only test files were touched, and the pattern is largely the same in each. The goal here is to move the codebase to more modern testing patterns while maintaining compatibility with the platforms we support. Additionally, the vast majority of the line changes are removing indentation that's no longer necessary.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,0 +1,18 @@
+# Copyright 2026 Open Source Robotics Foundation, Inc.
+# Licensed under the Apache License, Version 2.0
+
+from pathlib import Path
+
+import pytest
+
+
+if getattr(pytest, 'version_tuple', ()) < (3, 9):
+    @pytest.fixture
+    def tmp_path(tmpdir):
+        """
+        Compatibility fixture for temporary directory allocation.
+
+        This can be removed when we drop support for platforms with Pytest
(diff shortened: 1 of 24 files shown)


KmoM88 (approved):
LGTM. Great cleanup. Do you know if there are any other testing patterns that could be unified in a similar way in other projects?

cottsay:
> Do you know if there are any other testing patterns that could be unified in a similar way in other projects?

Great question. This particular pattern is a great candidate and could probably be cleaned up in a lot of other colcon projects.

In general, some of the simpler colcon extensions (for example, [colcon-defaults](https://github.com/colcon/colcon-defaults)) are missing tests entirely. A lot of these packages can usually copy the testing approach taken in similar components in `colcon-core` and just tweak them for the specific implementation.

PR #732, opened 2026-04-01
Implement interface for calling PEP 517 hooks

cottsay:
This implementation is copied nearly verbatim from the prototype `colcon_python_project` package.

Here's the spec for PEP 517: https://peps.python.org/pep-0517/

This is the central component for enabling colcon to act as a "build frontend" and the Python package build process will be built around this infrastructure. Additionally, non-[PEP 518](https://peps.python.org/pep-0518/) packages will extract package metadata during the identification and augmentation phases of the colcon workflow.

To help visualize where this fits in the overall standards-based Python build picture, this hook caller interface will invoke the appropriate backend hook (e.x. `build_wheel`) for a package to build a wheel file, and then colcon will install that wheel (after uninstalling any previous wheels for that package) into the install space.

The decoration infrastructure allows us to add backend-specific augmentations to hook invocations. Though this interface should allow colcon to support arbitrary backends, we'll implement decorators for "helping" the backends to support broader scenarios (like symlink installs) and optimize the overall process. For example, we'll need to augment `setuptools` builds in order to maintain feature parity with colcon's existing setuptools integration. We'll also use that decorator to move temporary artifacts out of the package source directory.

### DIFF
--- a/colcon_core/python_project/hook_caller/__init__.py
+++ b/colcon_core/python_project/hook_caller/__init__.py
@@ -1,0 +1,141 @@
+# Copyright 2022 Open Source Robotics Foundation, Inc.
+# Licensed under the Apache License, Version 2.0
+
+from contextlib import AbstractContextManager
+import os
+import pickle
+import sys
+
+from colcon_core.python_project.hook_caller import _call_hook
+from colcon_core.python_project.hook_caller import _list_hooks
+from colcon_core.python_project.spec import load_and_cache_spec
+from colcon_core.subprocess import run
+
+
+class _SubprocessTransport(AbstractContextManager):
+
+    def __enter__(self):
+        self.child_in, self.parent_out = os.pipe()
+        self.parent_in, self.child_out = os.pipe()
+
+        try:
+            import msvcrt
+        except ImportError:
+            os.set_inheritable(self.child_in, True)
+            self.pass_in = self.child_in
+            os.set_inheritable(self.child_out, True)
+            self.pass_out = self.child_out
+        else:
+            self.pass_in = msvcrt.get_osfhandle(self.child_in)
+            os.set_handle_inheritable(self.pass_in, True)
+            self.pass_out = msvcrt.get_osfhandle(self.child_out)
+            os.set_handle_inheritable(self.pass_out, True)
+
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        os.close(self.parent_out)
+        os.close(self.parent_in)
+        os.close(self.child_out)
+        os.close(self.child_in)
+
+
+class AsyncHookCaller:
+    """Calls PEP 517 style hooks asynchronously in a new process."""
+
+    def __init__(
+        self, backend_name, *, project_path=None, env=None,
+        stdout_callback=None, stderr_callback=None,
+    ):
+        """
+        Initialize a new AsyncHookCaller.
+
+        :param backend_name: The name of the PEP 517 build backend.
+        :param project_path: Path to the project's root directory.
+        :param env: Environment variables to use when invoking hooks.
+        :param stdout_callback: Callback for stdout from the hook invocation.
+        :param stderr_callback: Callback for stderr from the hook invocation.
+        """
+        self._backend_name = backend_name
+        self._project_path = str(project_path) if project_path else None
+        self._env = dict(env if env is not None else os.environ)
+        self._stdout_callback = stdout_callback
+        self._stderr_callback = stderr_callback
+
+    @property
+    def backend_name(self):
+        """Get the name of the backend to call hooks on."""
+        return self._backend_name
+
+    @property
+    def env(self):
+        """Get the environment variables to use when invoking hooks."""
+        return self._env
+
+    async def list_hooks(self):
+        """
+        Call into the backend to list implemented hooks.
+
+        This function lists all callable methods on the backend, which may
+        include more than just the hook names.
+
+        :returns: List of hook names.
+        """
+        args = [
+            sys.executable, _list_hooks.__file__,
+            self._backend_name]
+        process = await run(
+            args, None, self._stderr_callback,
+            cwd=self._project_path, env=self.env,
+            capture_output=True)
+        process.check_returncode()
+        hook_names = [
+            line.strip().decode() for line in process.stdout.splitlines()]
+        return [
+            hook for hook in hook_names if hook and not hook.startswith('_')]
+
+    async def call_hook(self, hook_name, **kwargs):
+        """
+        Call the given hook with given arguments.
+
+        :param hook_name: Name of the hook to call.
+        """
+        with _SubprocessTransport() as transport:
+            args = [
+                sys.executable, _call_hook.__file__,
+                self._backend_name, hook_name,
+                str(transport.pass_in), str(transport.pass_out)]
+            with os.fdopen(os.dup(transport.parent_out), 'wb') as f:
+                pickle.dump(kwargs, f)
+            have_callbacks = self._stdout_callback or self._stderr_callback
+            process = await run(
+                args, self._stdout_callback, self._stderr_callback,
+                cwd=self._project_path, env=self.env, close_fds=False,
+                capture_output=not have_callbacks)
+            process.check_returncode()
+            with os.fdopen(os.dup(transport.parent_in), 'rb') as f:
+                res = pickle.load(f)
+            return res
+
+
+def get_hook_caller(desc, **kwargs):
+    """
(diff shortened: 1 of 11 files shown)


test/test_flake8.py
```diff
@@ -25,7 +25,9 @@ def test_flake8():
         show_source=True,
     )
     style_guide_tests = get_style_guide(
-        extend_ignore=['D100', 'D101', 'D102', 'D103', 'D104', 'D105', 'D107'],
+        extend_ignore=[
+            'D100', 'D101', 'D102', 'D103', 'D104', 'D105', 'D106', 'D107',
+        ],
```
cottsay, line 30: I added `D106` to this list. Evidently we've never had a nested class definition in the test sources, but it makes sense to suppress the documentation requirement there as well.

colcon_core/python_project/hook_caller/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+        """
+        with _SubprocessTransport() as transport:
+            args = [
+                sys.executable, _call_hook.__file__,
+                self._backend_name, hook_name,
+                str(transport.pass_in), str(transport.pass_out)]
+            with os.fdopen(os.dup(transport.parent_out), 'wb') as f:
+                pickle.dump(kwargs, f)
+            have_callbacks = self._stdout_callback or self._stderr_callback
+            process = await run(
```
KmoM88, line 123: Question about the performance of the process-per-hook design, specifically in the new `AsyncHookCaller.call_hook` method in `colcon_core/python_project/hook_caller/__init__.py`.
The pattern of spawning a new process for isolation is similar, as far as I understand, to how colcon already handles `setup.py`. However, PEP 517 build can involve more, smaller hook calls per package.
Given that the number of subprocess invocations per package could increase, is there any concern this might become a noticeable performance issue in very large workspaces? Just curious about your thoughts on the scalability of this pattern.

colcon_core/python_project/hook_caller/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+                args, self._stdout_callback, self._stderr_callback,
+                cwd=self._project_path, env=self.env, close_fds=False,
+                capture_output=not have_callbacks)
+            process.check_returncode()
+            with os.fdopen(os.dup(transport.parent_in), 'rb') as f:
+                res = pickle.load(f)
```
KmoM88: I think there is a potential deadlock issue here related to IPC handling within the event loop.

Because `pickle.dump` is synchronous, blocking call, writing to the file descriptor can block the entire event loop if the payload exceeds the OS pipe buffer.

While this might not affect basic use cases with small payloads, it is possible to freeze in complex bigger scenarios (large dependency list or extensive config_settings).

Has a fully non-blocking IPC approach using asyncio streams been considered to make this more robust for advanced use cases? Maybe this is over-optimizing and can be considered if an issue is found and asyncio would remove the low level OS pipes from the previous comment.

colcon_core/python_project/hook_caller/__init__.py
```diff
@@ -0,0 +1,141 @@
...
+import pickle
+import sys
+
+from colcon_core.python_project.hook_caller import _call_hook
+from colcon_core.python_project.hook_caller import _list_hooks
+from colcon_core.python_project.spec import load_and_cache_spec
+from colcon_core.subprocess import run
+
+
+class _SubprocessTransport(AbstractContextManager):
```
KmoM88, line 16: I'm curious about the `_SubprocessTransport` class here. Manually handling low-level OS pipes and platform-specific inheritance logic (like `msvcrt`) can be brittle. In `colcon_core/subprocess.py` the subprocess management is done using `asyncio.create_subprocess_exec` if I'm not wrong. Wouldn't be posible to simplify the IPC by using asyncio? Leveraging `asyncio.create_subprocess_exec` with `stdin=asyncio.subprocess.PIPE` and `stdout=asyncio.subprocess.PIPE` directly could align better with existing implementation and address the possible deadlock concern in one of the next comments.

colcon_core/python_project/hook_caller/__init__.py
KmoM88: About the use of `pickle` for deserializing the hook's return value.  I understand that PEP 517 operates on a trust model, but there is still a potential for remote code execution if a compromised  build backend returns a malicious payload. Given that the standard PEP 517 hooks return simple, serializable data types (like  strings and lists), would it be an improvement to use a safer serialization format?
Using the `json` library would completely mitigate this risk without adding new dependencies. Alternatively,  if more complex objects are a future concern, a library like `jsonpickle` could also work (though it would add a new dependency).  This might be a great way to make the IPC mechanism even more robust.  What are your thoughts on this? Is something we should worry about at the moment?

KmoM88:
This PR provides the low-level AsyncHookCaller tool. For the high-level integration, is the plan to develop PEP 517 package support within the [`colcon-python-project`](https://github.com/colcon/colcon-python-project) repository, making it a functional extension similar to [`colcon-python-setup-py`](https://github.com/colcon/colcon-python-setup-py)?

KmoM88 (commented):
LGTM overall. This are just some comments to discuss and get more context about this feature. Great Job!

colcon_core/python_project/hook_caller/__init__.py
cottsay: This is a great question. The problem with using stdin/stdout is that PEP 517 doesn't forbid backends from using stdout for normal console messages, so we can't use it as a data pipeline. The purpose of this class is really just to set up additional pipes which behave very similar to stdin/stdout already do, but with a dedicated purpose.

colcon_core/python_project/hook_caller/__init__.py
cottsay: It is absolutely a problem, and one of my biggest critiques of PEP 517. In the worst case, we may actually need to build a project just to determine what the name of the package is. :(

Still, a worst-case scenario will only make three subprocess invocations during a typical colcon operation:
1. During package identification
2. During package build to detect if editable installs can be performed by the backend
3. During package build to actually build the wheel

In reality, we'll rarely do (1). For ROS packages, we'll have `package.xml` to facilitate metadata discovery. For (2), we could probably move to a fallback-style behavior where an editable install is attempted without first looking for the hooks and then ensure that it's communicated properly back to colcon that the hook failed because it doesn't exist, so that colcon can fall back to a traditional wheel instead of an editable one.

tl;dr - process invocations are indeed slow, but I think we can mitigate the fallout. In the end, it's more important to conform to PEP 517:
> Frontends should call each hook in a fresh subprocess, so that backends are free to change process global state (such as environment variables or the working directory). A Python library will be provided which frontends can use to easily call hooks this way.

colcon_core/python_project/hook_caller/__init__.py
cottsay: Hmm, that's a good point. I see some options to consider:
1. Do the synchronous pickling separately from writing to the socket, the latter of which could be done asynchronously
2. Move the pickle-and-write operation to happen after we attempt to spawn the subprocess
3. Put a `BufferedWriter` between the pickle operation and the pipe to the subprocess.

Thoughts?

cottsay: Looking at the hooks defined today, they all seem to accept and return only Python primitives. I'm willing to bet that Pickle will be more performant, but we can't limit Pickle to only Python primitives, so switching to another serializer will give us that security (indirectly).

I'll give JSON a shot.

cottsay (commented):
Thanks for giving this such a thorough look!

colcon_core/python_project/hook_caller/__init__.py
cottsay: Updated to use JSON in fb31156. It was a pretty clean change.

cottsay: After moving from Pickle to JSON, the read/write operations got separated from the (de)serialization anyway, so I went ahead and did (1) in 70c8019.

cottsay: Oof, actually it looks like that won't work on Windows. Reverted.

cottsay: I just pushed the write to a background thread in 0727c07, which should improve this scenario.

colcon_core/python_project/hook_caller/__init__.py
KmoM88: Oh, I see. Thank you for the clarification.

KmoM88 (approved):
Thanks so much for your patience and for being so thorough in addressing all the feedback.
No more comments on my side, LGTM. Happy to approve and continue with next phases to adopt PEP 517.

PR #738, opened 2026-06-15
Re-work extension point test mocking

cottsay:
These brittle mocks stopped working right in Python 3.15. Rather than mock the functions, we can mock the data and let normal Python metadata discovery happen.

Changes to the test functions are probably best viewed with whitespace modifications hidden due to the indentation changes in the test body: https://github.com/colcon/colcon-core/pull/738/changes?w=1


cottsay:
This _might_ be a good change to make long-term, but the bug that I'm trying to fix is actually a lot simpler, so I'm going to punt this for now.

Closing in favor of #739.

PR #743, opened 2026-08-17
Improve argument coloring where available

cottsay:
* Use %(default)s where appropriate
* Otherwise, use backticks around default values
* Use backticks when referring to console commands, such as references to other command line options or example command invocations

Example with Python 3.15:
out

Note that we're not actually choosing the palette here, we're just giving `argparse` the information it needs to align with other console output.

### DIFF
--- a/colcon_core/command.py
+++ b/colcon_core/command.py
@@ -19,8 +19,8 @@ from colcon_core.environment_variable import EnvironmentVariable
 """Environment variable to set the warnings filter for colcon modules"""
 WARNINGS_ENVIRONMENT_VARIABLE = EnvironmentVariable(
     'COLCON_WARNINGS',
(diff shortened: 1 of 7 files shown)


PiousCrossten (approved):
Works fine, nothing breaks. Ran the tests locally, flake8 and spell check pass, coverage unchanged.

One thing that looks like it slipped through:
 verb/test.py:93 and :98 - the --build-base / --install-base args there are identical to the ones in verb/build.py (same wrap_default_value() defaults), but they still say (default: build) / (default: install). **So colcon build --help gets the coloring and colcon test --help doesn't.**

 Rest LGTM 👍

cottsay:
Ah, well caught @PiousCrossten! I've addressed this in b2008de.

PR #744, opened 2026-08-18
Add and use new `EnvironDict` class

cottsay:
On Win32, the `os.environ` mapping behaves in many case-insensitive ways. When we copy `os.environ` to a typical `dict` instance, we lose those behaviors.

This new class preserves the case insensitivity on Windows and behaves like a normal OrderedDict on non-Windows platforms.

~~In draft for now, because I'm still trying to understand the exact behaviors that Windows exhibits. In particular, it might be closer in behavior to just convert all keys to UPPERCASE before accessing the underlying dict.~~

Okay, from what I can tell, Python converts all keys in `os.environ` to UPPERCASE on Windows, but since colcon is capturing the environment variables using `set`, we don't see the same treatment. If the goal was to align `EnvironDict` with `os.environ` on Windows precisely, we'd want to convert everything to UPPERCASE as it seems to implicitly do. However, given that we want this class to be used on the variables we capture with `set` as well, I think we should just preserve the original case but make access case-insensitive.

Assisted-by: Gemini 3.5 Flash

> [!NOTE]
> The Windows implementation of `EnvironDict` is a lot bigger than it should be. Implementing `Mapping` was a lot simpler, but the public colcon API specifically returns an instance of `dict`, so we need the new class to also inherit from `dict` to maintain compatibility.


cottsay:
Tapping Pepper for review here (via @claraberendsen)

test/test_environment_variable.py
```diff
@@ -0,0 +1,80 @@
+# Copyright 2026 Dirk Thomas
```
knmcguire, line 1: Do you still want to have Dirk's name here or your own?

knmcguire (approved):
lgtm! and your explanation makes sense

The one thing perhaps to mention is that 'move_to_end()' from OrderedDict is not changed and can't handle the case-sensitive cases. It's not used anywhere though... so... I don't think it has consequences now?

fyi, did you use any agents here? It's not mentioned in the commits

test/test_environment_variable.py
cottsay: Ah, this should have been OSRF. Fixed, thanks!

cottsay:
> 'move_to_end()' from OrderedDict is not changed 

An oversight, addressed in the latest change.

> did you use any agents here? It's not mentioned in the commits

I did and I definitely forgot to mention it. Good call out.

PR #746, opened 2026-09-17
Add 'printed_strlen' utility function

cottsay:
This function is useful in output styling contexts where we need to determine how many columns a stylized string will occupy.

Inspired by @christophebedard : https://github.com/colcon/colcon-notification/pull/80#discussion_r4030842844


christophebedard (approved):
🚀

